PR #49, opened 2024-01-11
deps: Lockdown typebox peer dependency to patch

masad-frost:
Typebox is a peer dependency because we want people to author their own types. Typebox is v0.x.x which means it can break at any change.

https://github.com/sinclairzx81/typebox/pull/689 0.32.0 seems to be a breaking change.

I chose to just lock down minor changes since that seems to match what typebox maintainer abides by..

If locking down the version becomes an issue we can make typebox a direct dependency and re-export it for consumers (but that kinda sucks)

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,6 @@
     "ws": "^8.13.0"
   },
   "devDependencies": {
-    "@sinclair/typebox": "~0.31.28",
     "@types/ws": "^8.5.5",
     "@vitest/ui": "^1.1.0",
     "prettier": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -57,11 +57,12 @@
     "nanoid": "^4.0.2"
   },
   "peerDependencies": {
-    "@sinclair/typebox": "^0.31.28",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -57,11 +57,12 @@
     "nanoid": "^4.0.2"
   },
   "peerDependencies": {
-    "@sinclair/typebox": "^0.31.28",
+    "@sinclair/typebox": "~0.31.28",
     "isomorphic-ws": "^5.0.0",
     "ws": "^8.13.0"
   },
   "devDependencies": {
+    "@sinclair/typebox": "~0.31.28",
```
jackyzha0, line 65: we can remove from `devDependencies` i think, `npm` default behaviour automatically installs `peerDependencies`: https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies

jackyzha0 (approved):

package.json
masad-frost: Ah I'm old school

PR #51, opened 2024-02-06
protocol: frame messages with length-prefix instead of newlines

jackyzha0:
## Why

We want to have newlines in our messages :) The old approach failed when we used the binary codec with a string or Uint8Array that contained a newline in it. This is because we framed our messages with newlines so we would get cases where a single message would be split in two.

## What changed

Use `uint32`-prefix framing instead.

### DIFF
--- a/transport/transforms/messageFraming.test.ts
+++ b/transport/transforms/messageFraming.test.ts
@@ -1,14 +1,14 @@
-import { Delimiter } from './delim';
+import { MessageFramer } from './messageFraming';
 import { describe, test, expect, vi } from 'vitest';
 
-describe('Uint32LengthEncodedDelimiter', () => {
+describe('MessageFramer', () => {
   const encodeMessage = (message: string) => {
-    return Delimiter.writeDelimited(Buffer.from(message));
+    return MessageFramer.write(Buffer.from(message));
   };
 
   test('basic transform', () => {
     const spy = vi.fn();
-    const parser = Delimiter.createDelimitedStream();
+    const parser = MessageFramer.createFramedStream();
 
     parser.on('data', spy);
     parser.write(encodeMessage('content 1'));
@@ -26,7 +26,7 @@ describe('Uint32LengthEncodedDelimiter', () => {
 
   test('handles partial messages across chunks', () => {
     const spy = vi.fn();
-    const parser = Delimiter.createDelimitedStream();
+    const parser = MessageFramer.createFramedStream();
 
     const msg = encodeMessage('content 1');
     const part1 = msg.subarray(0, 5); // Split the encoded message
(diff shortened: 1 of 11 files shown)


transport/transforms/messageFraming.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    expect(spy).toHaveBeenCalledTimes(0);
+    expect(err).toHaveBeenCalledTimes(1);
+  });
+
+  test('consistent byte length calculation with emojis and unicode', () => {
+    const parser = MessageFramer.createFramedStream();
+    const spy = vi.fn();
+    parser.on('data', spy);
+
+    const emojiMessage = '🇧🇪🇨🇦👨‍👩‍👧‍👦';
```
lhchavez, line 113: giuseppe would be proud.

transport/transforms/messageFraming.ts
```diff
@@ -0,0 +1,91 @@
+import { Transform, TransformCallback, TransformOptions } from 'node:stream';
+
+export interface LengthEncodedOptions extends TransformOptions {
+  /** Maximum in-memory buffer size before we throw */
+  maxBufferSizeBytes: number;
+  preAllocatedBufferSize: number;
+}
+
+/**
+ * A transform stream that emits data each time a message with a uint32 length prefix is received.
```
lhchavez, line 10:
```suggestion
 * A transform stream that emits data each time a message with a network/BigEndian uint32 length prefix is received.
```

transport/transforms/messageFraming.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function createLengthEncodedStream(options?: Partial<LengthEncodedOptions>) {
+  return new Uint32LengthPrefixFraming({
+    maxBufferSizeBytes: options?.maxBufferSizeBytes ?? 16 * 1024 * 1024, // 16MB
+    preAllocatedBufferSize: options?.preAllocatedBufferSize ?? 16 * 1024, // 16KB
+  });
+}
+
+export const MessageFramer = {
```
lhchavez, line 84: is it worth having a no-op framer for when we write these things to a websocket transport (or something that already has framing built-in? like a SEQPACKET UDS)

lhchavez (approved):

transport/transforms/messageFraming.ts
lhchavez: ah nevermind, i saw that those cases don't even use one. carry on.

transport/transforms/messageFraming.test.ts
jackyzha0: the notorious 🇩🇪 returns

gburtini: :heart:

PR #52, opened 2024-02-20
[protocol] distinguish between connections and sessions

jackyzha0:
## Why

Part 1 of improving River reconnect strategy. 

- The way we re-send buffered messages is really naive. It doesn’t take into account the possibility of the server losing state between reconnects.
- We need a way to know when the client reconnects to a different session so it knows how to reset its own state and clear its sendBuffer and sendQueue

This PR introduces the concept of a session, distinct from a single connection.

- A connection is the actual raw underlying transport connection
  - It’s responsible for dispatching to/from the actual connection itself
  - This should be instantiated as soon as the client/server has a connection
  - It’s tied to the lifecycle of the underlying transport connection (i.e. if the WS drops, this connection should be deleted)
- A session is a higher-level abstraction that operates over the span of potentially multiple transport-level connections
  - It’s responsible for tracking any metadata for a particular client that might need to be persisted across connections (i.e. the sendBuffer and sendQueue)
  - This will only be considered disconnected if the server tells the client that we’ve reconnected but it doesn’t recognize us anymore (server definitely died) or we hit a grace period after a connection disconnect
    - (this PR doesn't include the transparent reconnect but connected to diff server session check yet but will add that to the ack refactor PR)

We then define two types of reconnects:
1. Transparent reconnects: when we know the network caused some disconnection but the underlying state on the server is still intact (i.e. a connection is dropped and reestablished but the session is the same). This will preserve the sendBuffer so to the application, it shouldn’t know a disconnect even happened. If a transparent reconnect fails after a certain grace period (defaults to 3s in the protocol but is adjustable), we give up on the connection and consider it a hard reconnect
2. Hard reconnect: when we reconnected but the server has lost state and everything needs to be reinitialized (reset-the-world). This will clear the sendBuffer so we don’t send stale messages

## What Changed
- Adds a `sessionStatus` event listener alongside the `connectionStatus` event listener with appropriate tests
- UDS now has it's own connection class so it can clean up appropriately
  - rename `unixsocket` to `uds`
- stdio now has a client/server implementation to distinguish responsibility of who connects first
- Improve fake-timers and general test stability
  - Improve how strict post-test checks are in `cleanup.test.ts`
  - Add `./flake.sh` as a flake detector
- Various bug fixes
  - Shouldn't send close message after we get a hard disconnect
  - Moved reconnection logic outside of WS and into transport so it's transport agnostic :)
- Reconnect now happens after onDisconnect rather than on failed send
- Fixed up some incorrect logs and annotate connections and sessions in logs


transport/transport.ts
```diff
@@ -306,38 +422,43 @@ export abstract class Transport<ConnType extends Connection> {
...
-    let conn = this.connections.get(msg.to);
+    let session = this.sessions.get(msg.to);
+    if (!session) {
+      session = this.createSession(msg.to, undefined);
+      log?.info(
+        `${this.clientId} -- no session for ${msg.to}, created a new one (id: ${session.id})`,
+      );
+    }
+
+    let conn = session?.connection;
```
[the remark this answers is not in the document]
masad-frost: we should probably add some of the eslint rules that use tsserver.

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
     this.clientId = clientId;
     this.state = 'open';
+    this.inflightConnectionPromises = new Map();
   }
 
   /**
    * Abstract method that creates a new {@link Connection} object.
-   * This should call {@link onConnect} when the connection is established.
+   * This should call {@link onConnect} when the connection is established
+   * and {@link onDisconnect} when the connection is closed.
```
masad-frost, line 133: Maybe I'm missing something, but why is the implementer responsible for calling `onConnect` and `onDisconnect`? Why don't they just call `connect` and which calls the implemented `createNewOutgoingConnection` and returns a `Connection`. Then inside `connect` we use the returned connection and pass it to `onConnect` and `onDisconnect`, this would probably require us to add more event listeners to the `Connection` interface like i mentioned in another comment.

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
+    let reconnectPromise = this.inflightConnectionPromises.get(to);
+    if (!reconnectPromise) {
+      reconnectPromise = this.createNewOutgoingConnection(to);
+      this.inflightConnectionPromises.set(to, reconnectPromise);
+    }
+
+    try {
+      const conn = await reconnectPromise;
+      if (this.state !== 'open') {
+        this.inflightConnectionPromises.delete(to);
```
masad-frost, line 170: shouldn't we delete this regardless of the state after the promise returns? Maybe I'm missing something and we need to keep this around when the state is open?

transport/transport.ts
```diff
@@ -306,38 +422,43 @@ export abstract class Transport<ConnType extends Connection> {
           msg,
         )}`,
       );
-      return msg.id;
+      return false;
     }
 
-    let conn = this.connections.get(msg.to);
+    let session = this.sessions.get(msg.to);
+    if (!session) {
```
masad-frost, line 429: hmm, shouldn't we have called `createSession` in `onConnect`?

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
+    session.cancelGrace();
+
+    // if there are any unacked messages in the sendQueue, send them now
+    for (const id of session.sendQueue) {
+      const msg = session.sendBuffer.get(id);
+      if (msg) {
+        const ok = this.send(msg);
+        if (!ok) {
+          // this should never happen unless the transport has an
+          // incorrect implementation of `createNewConnection`
```
masad-frost, line 256:
```suggestion
          // incorrect implementation of `createNewOutgoingConnection`
```

transport/transport.ts
```diff
@@ -306,38 +422,43 @@ export abstract class Transport<ConnType extends Connection> {
           msg,
         )}`,
       );
-      return msg.id;
+      return false;
```
masad-frost, line 425: what happens to the message in this case?

transport/transport.ts
```diff
@@ -306,38 +422,43 @@ export abstract class Transport<ConnType extends Connection> {
...
-        return msg.id;
-      }
+      if (ok) return true;
+      log?.info(
+        `${this.clientId} -- failed to send on connection (id: ${conn.id}) to ${msg.to}, queuing msg ${msg.id}`,
+      );
+    } else {
+      log?.info(
+        `${this.clientId} -- connection to ${msg.to} doesn't exist, queuing msg ${msg.id}`,
+      );
```
masad-frost, line 454: i feel like i asked about this before, but if the message doesn't require an ack, where is it queued up? The `sendQueue` only has ids

masad-frost (commented):
Mostly questions and suggestions, still wrapping my head around all of this.

bradymadden97:
I think Faris had better suggestions than me. I had a really hard time reviewing this PR. I'm struggling to keep all the concepts in my head - and we keep adding new ones - as well as understand the control flow. I saw the diagram going from client -> transport -> session -> connection. But just having a difficult time understanding the concepts when they're in code. I think the class structure is making it more difficult - as we have implementations of things in many different places, it's unclear what layer is responsible for what. 

Like the transport base class will have a bunch of important logic around sessions & connections, but then it will defer some of that logic, via abstract methods, to individual impl classes. But then those impl classes will sometimes call back into a shared method from the base class. It's hard to keep that straight.

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
-      this.send(msg);
+    // otherwise, this is a duplicate session from the same user, let's consider
+    // the old one as dead and call this one canonical
+    this.closeStaleConnectionForSession(conn, session);
+    session.connection = conn;
+    session.cancelGrace();
+
+    // if there are any unacked messages in the sendQueue, send them now
+    for (const id of session.sendQueue) {
+      const msg = session.sendBuffer.get(id);
```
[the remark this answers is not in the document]
jackyzha0: this is buggy and this fix is still buggy, will fix this in the 3rd PR of the stack which addresses the underlying ack implementation

transport/session.ts
```diff
@@ -0,0 +1,135 @@
...
+ * connection      1---------3   1-------3
+ *                                       ^-^ grace period
+ *             ^---^ session is created as soon
+ *                   as we send an outgoing message
+ * ```
+ *
+ * 1. connectionStatus :: connect
+ * 2. sessionStatus    :: connect
+ * 3. connectionStatus :: disconnect
+ * 4. sessionStatus    :: disconnect
```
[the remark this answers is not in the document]
jackyzha0: oof yeah hm I wonder how we can make this more clear, what I'm trying to illustrate is the lifetime of a "session" vs a "connection" on both the server and the client because the lifetimes are slightly different

1. corresponds to a connection status event with type connect
2. corresponds to a session status event with type connect
3. corresponds to a connection status event with type disconnect
4. corresponds to a session status event with type disconnect

the arrows try to explain some of the gaps between the two and the various reasons

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
 
   /**
    * The downstream implementation needs to call this when a connection is closed.
    * @param conn The connection object.
    */
-  onDisconnect(conn: ConnType) {
-    log?.info(`${this.clientId} -- disconnect from ${conn.connectedTo}`);
-    conn.close();
-    this.connections.delete(conn.connectedTo);
+  onDisconnect(conn: ConnType, connectedTo: TransportClientId | undefined) {
```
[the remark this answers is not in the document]
jackyzha0: this gave me slightly better compiler hints when i was refactoring (i needed to explicitly pass in undefined rather than just leaving it as the one-argument call it was before) -- we should always try to pass in `connectedTo` if we have the information

transport/session.ts
```diff
@@ -0,0 +1,135 @@
...
+  id: string;
+
+  /**
+   * A timeout that is used to close the session if the connection is not re-established
+   * within a certain period of time.
+   */
+  private graceExpiryTimeout?: ReturnType<typeof setTimeout>;
+
+  constructor(connectedTo: TransportClientId, conn: ConnType | undefined) {
+    this.id = `sess-${unsafeId()}`; // for debugging, no collision safety needed
```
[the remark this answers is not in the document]
masad-frost: This value has meaning afaik

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
 
-      this.send(msg);
+    // otherwise, this is a duplicate session from the same user, let's consider
+    // the old one as dead and call this one canonical
+    this.closeStaleConnectionForSession(conn, session);
+    session.connection = conn;
+    session.cancelGrace();
+
+    // if there are any unacked messages in the sendQueue, send them now
+    for (const id of session.sendQueue) {
```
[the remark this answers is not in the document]
jackyzha0: yup thats actually probably more correct, this is refactored out in the ack rewrite anyways though

transport/session.ts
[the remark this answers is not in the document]
masad-frost: Ok I get it now, I think spelling it out is probably more understandable than the diagram. The PR description is good to plop in here

transport/session.ts
```diff
@@ -0,0 +1,135 @@
...
+  id: string;
+  constructor() {
+    this.id = `conn-${unsafeId()}`; // for debugging, no collision safety needed
+  }
+
+  /**
+   * Handle adding a callback for when a message is received.
+   * @param msg The message that was received.
+   */
+  abstract onData(cb: (msg: Uint8Array) => void): void;
```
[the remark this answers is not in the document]
jackyzha0: talked in person, let's move that to https://github.com/replit/river/pull/53

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
-        continue;
-      }
+    log?.info(
+      `${this.clientId} -- new connection (id: ${conn.id}) for existing session (id: ${session.id}) to ${connectedTo}`,
+    );
 
-      this.send(msg);
+    // otherwise, this is a duplicate session from the same user, let's consider
+    // the old one as dead and call this one canonical
+    this.closeStaleConnectionForSession(conn, session);
```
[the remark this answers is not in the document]
jackyzha0: consider we have some connection A from Bob that has died but hasn't hit the grace period where we removed it

Bob could have tried to make a new connection B before we hit the grace period so we need to consider A dead so we don't break the 1 session <-> 1 connection invariant

transport/transport.ts
jackyzha0: yup, will do that in pr 2/3! (https://github.com/replit/river/pull/53/files)

transport/transport.ts
jackyzha0: this allows us to cache successful connection requests so that subsequent .connect calls can reuse it until it we know for sure that it is unhealthy

transport/transport.ts
jackyzha0: hmm yeah you're right, this invariant case should never happen, lemme test without it

jackyzha0: actually no consider the client case, we can send before we have a session to the server so this starts the session

transport/transport.ts
jackyzha0: silently discarded, closed transports are not meant to be reopened

transport/transport.ts
jackyzha0: in this version of the protocol, all messages that are not acks require acks (and hence the bug that acks themselves might accidentally be dropped, we'll fix this in the 3rd pr of the stack)

router/server.ts
```diff
@@ -76,7 +77,7 @@ class RiverServer<Services extends ServiceDefs> {
     this.streamMap = new Map();
     this.clientStreams = new Map();
     this.transport.addEventListener('message', this.handler);
-    this.transport.addEventListener('connectionStatus', this.onDisconnect);
+    this.transport.addEventListener('sessionStatus', this.onDisconnect);
```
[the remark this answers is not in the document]
jackyzha0: `sessionStatus` can either be connect or disconnect, I fixed the naming

transport/session.ts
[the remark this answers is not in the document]
jackyzha0: it's actually safe to change, we don't key anything on session or connection id

transport/transport.ts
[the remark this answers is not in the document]
jackyzha0: responded to faris but yeah this is buggy but buggy in the same way it was before, we're fixing this with the ack upgrades

transport/session.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * An array of message IDs that are waiting to be sent over the connection.
+   * This builds up if the there is no connection for a period of time.
+   */
+  sendQueue: Array<MessageId>;
+
+  /**
+   * The buffer of messages that have been sent but not yet acknowledged.
+   */
+  sendBuffer: Map<MessageId, OpaqueTransportMessage>;
```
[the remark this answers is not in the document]
jackyzha0: yes I agree, see previous message!

transport/transport.ts
```diff
@@ -12,40 +11,28 @@ import {
...
-  transport: Transport<Connection>;
-
-  constructor(
-    transport: Transport<Connection>,
-    connectedTo: TransportClientId,
-  ) {
-    this.connectedTo = connectedTo;
-    this.transport = transport;
-  }
+export type TransportStatus = 'open' | 'closed' | 'destroyed';
```
lhchavez, line 17: can there be a lil docstring explaining each status?

transport/transport.ts
```diff
@@ -12,40 +11,28 @@ import {
...
-  abstract send(msg: Uint8Array): boolean;
-  abstract close(): void;
+export interface TransportOptions {
+  retryIntervalMs: number;
+  retryAttemptsMax: number;
+  codec: Codec;
 }
 
-export type TransportStatus = 'open' | 'closed' | 'destroyed';
+export const DEFAULT_WS_RETRY_INTERVAL_MS = 250;
```
lhchavez, line 25: can there be a lil docstring explaining what "retry interval" means?

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
    */
-  abstract createNewConnection(to: TransportClientId): Promise<void>;
+  protected abstract createNewOutgoingConnection(
+    to: TransportClientId,
+  ): Promise<ConnType>;
 
   /**
-   * The downstream implementation needs to call this when a new connection is established.
-   * @param conn The connection object.
+   * Manually attempts to connect to a client.
```
lhchavez, line 150: Is there another function that automatically attempts to connect? (in other words, why is there a "manually" qualifier here? might be worth expanding upon the consequences  of or recommendations for calling this.)

might also be worth calling out what the preconditions to be able to call this are (the state not being open AND `tryReconnecting` being set to false).

transport/transport.ts
```diff
@@ -90,111 +77,238 @@ export abstract class Transport<ConnType extends Connection> {
...
+      this.state = 'open';
+    } catch (err: unknown) {
+      // retry on failure
+      this.inflightConnectionPromises.delete(to);
+      if (attempt >= this.options.retryAttemptsMax) {
+        throw new Error(
+          `${this.clientId} -- connection to ${to} failed after ${attempt} attempts (${err}), giving up`,
+        );
+      } else {
+        // linear backoff
```
lhchavez, line 184: can we do exponential with jitter? i'm worried about stampedes.

transport/transport.ts
lhchavez: on this thread, might be worth persisting this bit of lore as a comment in code for future spelunkers.

transport/transport.ts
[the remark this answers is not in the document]
lhchavez: lol three comments. might be worth adding a lil `TODO: this is broken, but it's just a refactor, will be fixed in the next commit` for future spelunkers (if we ever need to bisect)

transport/transport.ts
```diff
@@ -211,25 +325,23 @@ export abstract class Transport<ConnType extends Connection> {
...
-        ...parsedMsg,
-        serviceName:
-          parsedMsg.serviceName === null ? undefined : parsedMsg.serviceName,
-        procedureName:
-          parsedMsg.procedureName === null
-            ? undefined
-            : parsedMsg.procedureName,
-      };
-    } else {
+    if (!Value.Check(OpaqueTransportMessageSchema, parsedMsg)) {
```
lhchavez, line 328: guard clause pattern ftw!

transport/transport.ts
lhchavez: can this convo be persisted as a code comment?

lhchavez (commented):
ok i think i got the first iteration in \o/ will do a second pass tomorrow

__tests__/typescript-stress.test.ts
```diff
@@ -1,14 +1,14 @@
...
+import { OpaqueTransportMessage } from '../transport/message';
 import { createServer } from '../router/server';
-import { Connection, Transport } from '../transport/transport';
-import { NaiveJsonCodec } from '../codec/json';
+import { Transport } from '../transport/transport';
 import { createClient } from '../router/client';
 import { Ok } from '../router/result';
 import { buildServiceDefs } from '../router/defs';
 import { TestServiceConstructor } from './fixtures/services';
+import { Connection } from '../transport';
```
masad-frost, line 11:
```suggestion
import { Transport, Connection } from '../transport/transport';
import { createClient } from '../router/client';
import { Ok } from '../router/result';
import { buildServiceDefs } from '../router/defs';
import { TestServiceConstructor } from './fixtures/services';
```

masad-frost (approved):
other than things we follow up in other PRs and Luis's comments LGTM.

PR #53, opened 2024-02-20
[protocol] Pre-session handshake

jackyzha0:
## Why

Having a pre-session handshake to exchange information is useful! Currently, we do this implicitly on the first message of the connection but explicit good 👍 

## What changed

- Separated out client/server transport into sub-abstract classes of transport to reduce boilerplate in sending/receiving handshake sequence
- Implement the pre-session handshake (connection isn't turned into session until the handshake is completed)

### DIFF
--- a/router/context.ts
+++ b/router/context.ts
@@ -18,9 +18,7 @@
  * }
  * ```
  */
-export interface ServiceContext {
-  state: object | unknown;
-}
+export interface ServiceContext {}
 
 /**
  * The {@link ServiceContext} with state. This is what is passed to procedures.
--- a/transport/impls/stdio/client.ts
+++ b/transport/impls/stdio/client.ts
@@ -36,14 +36,7 @@ export class StdioClientTransport extends ClientTransport<StreamConnection> {
   async createNewOutgoingConnection(to: TransportClientId) {
     log?.info(`${this.clientId} -- establishing a new stream to ${to}`);
     const conn = new StreamConnection(this.input, this.output);
-    conn.addDataListener(this.receiveWithBootSequence(conn));
-    const cleanup = () => {
-      this.onDisconnect(conn, to);
-      this.connect(to);
-    };
-
-    this.input.addListener('close', cleanup);
-    this.output.addListener('close', cleanup);
+    this.handleConnection(conn, to);
     return conn;
   }
 }
--- a/transport/impls/stdio/client.ts
+++ b/transport/impls/stdio/client.ts
@@ -1,6 +1,6 @@
 import { log } from '../../../logging';
 import { TransportClientId } from '../../message';
-import { Transport, TransportOptions } from '../../transport';
+import { ClientTransport, TransportOptions } from '../../transport';
 import { StreamConnection } from './connection';
 
 /**
@@ -8,7 +8,7 @@ import { StreamConnection } from './connection';
  * Will eagerly connect as soon as it's initialized.
  * @extends Transport
  */
-export class StdioClientTransport extends Transport<StreamConnection> {
+export class StdioClientTransport extends ClientTransport<StreamConnection> {
   input: NodeJS.ReadableStream = process.stdin;
   output: NodeJS.WritableStream = process.stdout;
   serverId: TransportClientId;
--- a/transport/impls/stdio/client.ts
+++ b/transport/impls/stdio/client.ts
@@ -36,7 +36,7 @@ export class StdioClientTransport extends ClientTransport<StreamConnection> {
   async createNewOutgoingConnection(to: TransportClientId) {
     log?.info(`${this.clientId} -- establishing a new stream to ${to}`);
     const conn = new StreamConnection(this.input, this.output);
-    conn.addDataListener((data) => this.handleMsg(this.parseMsg(data)));
+    conn.addDataListener(this.receiveWithBootSequence(conn));
     const cleanup = () => {
       this.onDisconnect(conn, to);
       this.connect(to);
@@ -44,7 +44,6 @@ export class StdioClientTransport extends ClientTransport<StreamConnection> {
 
     this.input.addListener('close', cleanup);
     this.output.addListener('close', cleanup);
-    this.onConnect(conn, to);
     return conn;
   }
 }
--- a/transport/impls/stdio/server.ts
+++ b/transport/impls/stdio/server.ts
@@ -1,6 +1,4 @@
-import { log } from '../../../logging';
 import { TransportClientId } from '../../message';
-import { Session } from '../../session';
 import { ServerTransport, TransportOptions } from '../../transport';
 import { StreamConnection } from './connection';
 
@@ -28,38 +26,7 @@ export class StdioServerTransport extends ServerTransport<StreamConnection> {
     super(clientId, providedOptions);
     this.input = input;
     this.output = output;
-
-    let session: Session<StreamConnection> | undefined = undefined;
-    const receiver = () => session?.connectedTo ?? 'unknown';
-
     const conn = new StreamConnection(this.input, this.output);
-    conn.addDataListener(
-      this.receiveWithBootSequence(conn, (establishedSession) => {
-        session = establishedSession;
-      }),
-    );
-
-    const cleanup = (session: Session<StreamConnection>) =>
-      this.onDisconnect(conn, session.connectedTo);
-
-    this.input.on('close', () => {
-      if (!session) return;
-      log?.info(
-        `${this.clientId} -- stream conn (id: ${
-          conn.debugId
-        }) to ${receiver()} disconnected`,
-      );
-      cleanup(session);
-    });
-
-    this.input.on('error', (err) => {
-      if (!session) return;
-      log?.warn(
-        `${this.clientId} -- error in stream connection (id: ${
-          conn.debugId
-        }) to ${receiver()}: ${err}`,
-      );
-      cleanup(session);
-    });
+    this.handleConnection(conn);
   }
 }
--- a/transport/impls/stdio/server.ts
+++ b/transport/impls/stdio/server.ts
@@ -62,10 +62,4 @@ export class StdioServerTransport extends ServerTransport<StreamConnection> {
       cleanup(session);
     });
   }
-
-  async createNewOutgoingConnection(to: string): Promise<StreamConnection> {
-    const err = `${this.clientId} -- failed to send msg to ${to}, client probably dropped`;
-    log?.warn(err);
-    throw new Error(err);
-  }
 }
--- a/transport/impls/stdio/server.ts
+++ b/transport/impls/stdio/server.ts
@@ -34,7 +34,7 @@ export class StdioServerTransport extends ServerTransport<StreamConnection> {
 
     const conn = new StreamConnection(this.input, this.output);
     conn.addDataListener(
-      this.receiveBootSequence(conn, (establishedSession) => {
+      this.receiveWithBootSequence(conn, (establishedSession) => {
         session = establishedSession;
       }),
     );
--- a/transport/impls/stdio/server.ts
+++ b/transport/impls/stdio/server.ts
@@ -1,7 +1,7 @@
 import { log } from '../../../logging';
 import { TransportClientId } from '../../message';
 import { Session } from '../../session';
-import { Transport, TransportOptions } from '../../transport';
+import { ServerTransport, TransportOptions } from '../../transport';
 import { StreamConnection } from './connection';
 
 /**
@@ -9,7 +9,7 @@ import { StreamConnection } from './connection';
  * This will sit idle until a client connects.
  * @extends Transport
  */
-export class StdioServerTransport extends Transport<StreamConnection> {
+export class StdioServerTransport extends ServerTransport<StreamConnection> {
   input: NodeJS.ReadableStream = process.stdin;
   output: NodeJS.WritableStream = process.stdout;
 
@@ -33,15 +33,11 @@ export class StdioServerTransport extends Transport<StreamConnection> {
     const receiver = () => session?.connectedTo ?? 'unknown';
 
     const conn = new StreamConnection(this.input, this.output);
-    conn.addDataListener((data) => {
-      const parsed = this.parseMsg(data);
-      if (!parsed) return;
-      if (!session && !this.connections.has(parsed.from)) {
-        session = this.onConnect(conn, parsed.from);
-      }
-
-      this.handleMsg(parsed);
-    });
+    conn.addDataListener(
+      this.receiveBootSequence(conn, (establishedSession) => {
+        session = establishedSession;
+      }),
+    );
 
     const cleanup = (session: Session<StreamConnection>) =>
       this.onDisconnect(conn, session.connectedTo);
--- a/transport/impls/uds/connection.ts
+++ b/transport/impls/uds/connection.ts
@@ -21,6 +21,14 @@ export class UdsConnection extends Connection {
     this.input.on('data', cb);
   }
 
+  addCloseListener(cb: () => void): void {
+    this.sock.on('close', cb);
+  }
+
+  addErrorListener(cb: (err: Error) => void): void {
+    this.sock.on('error', cb);
+  }
+
   send(payload: Uint8Array) {
     return this.sock.write(MessageFramer.write(payload));
   }
--- a/transport/impls/uds/connection.ts
+++ b/transport/impls/uds/connection.ts
@@ -21,6 +21,10 @@ export class UdsConnection extends Connection {
     this.input.on('data', cb);
   }
 
+  removeDataListener(cb: (msg: Uint8Array) => void): void {
+    this.input.off('data', cb);
+  }
+
   addCloseListener(cb: () => void): void {
     this.sock.on('close', cb);
   }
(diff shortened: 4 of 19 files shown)


router/context.ts
```diff
@@ -18,9 +18,7 @@
  * }
  * ```
  */
-export interface ServiceContext {
-  state: object | unknown;
-}
+export interface ServiceContext {}
```
jackyzha0, line 21: idk why this included state before, we have a separate type for this

transport/message.ts
```diff
@@ -37,16 +38,45 @@ export const TransportMessageSchema = <T extends TSchema>(t: T) =>
...
-export const ControlMessagePayloadSchema = Type.Object({
+/**
+ * Defines the schema for a transport close message. This is never constructed manually and is only
+ * used internally by the library for closing and cleaning up streams.
+ */
+export const ControlMessageCloseSchema = Type.Object({
   type: Type.Literal('CLOSE'),
 });
 
+export const PROTOCOL_VERSION = 'v1';
```
masad-frost, line 54: :heart:

transport/impls/stdio/client.ts
```diff
@@ -36,15 +36,14 @@ export class StdioClientTransport extends Transport<StreamConnection> {
   async createNewOutgoingConnection(to: TransportClientId) {
     log?.info(`${this.clientId} -- establishing a new stream to ${to}`);
     const conn = new StreamConnection(this.input, this.output);
-    conn.onData((data) => this.handleMsg(this.parseMsg(data)));
+    conn.onData(this.receiveWithBootSequence(conn));
     const cleanup = () => {
       this.onDisconnect(conn, to);
```
masad-frost, line 41: Mentioned in the other PR

https://github.com/replit/river/pull/52#discussion_r1499767224

masad-frost: > Maybe I'm missing something, but why is the implementer responsible for calling `onConnect` and `onDisconnect`? Why don't they just call `connect` and which calls the implemented `createNewOutgoingConnection` and returns a `Connection`. Then inside `connect` we use the returned connection and pass it to `onConnect` and `onDisconnect`, this would probably require us to add more event listeners to the `Connection` interface like i mentioned in another comment.

transport/impls/stdio/server.ts
```diff
@@ -33,15 +33,11 @@ export class StdioServerTransport extends Transport<StreamConnection> {
...
-        session = this.onConnect(conn, parsed.from);
-      }
-
-      this.handleMsg(parsed);
-    });
+    conn.addDataListener(
+      this.receiveWithBootSequence(conn, (establishedSession) => {
+        session = establishedSession;
+      }),
+    );
```
lhchavez, line 40: quick suggestion to [make invalid states unrepresentable](https://docs.google.com/document/d/1tdcHyhMy0tGI9JMosk6Gzzy_ItDRxZX-uIAa03YC7pU/edit?resourcekey=0-3GX7IW7Rsuy3QUcnP2bc5w): instead of adding the handshake as one of the multiple bits a message could have (what's the semantics of a message being both handshake and stream-open??), we can take the handshake out of the normal protocol messages and just parse it on its own (so we'd have `parseMsg` and `parseHandshakeMsg`).

which means that instead of having a `receiveWithBootSequence`, probably the play is to have a utility function:

```suggestion
    this.addDataListener(conn, (establishedSession) => {
      session = establishedSession;
    };
```

where `addDataListener` is something like this:

```typescript
  addDataListener(
    conn: ConnType,
    sessionCb?: (sess: Session<ConnType>) => void,
  ) {
    const firstMessageListener = (data) => {
      try {
        const parsed = this.parseHandshakeMsg(data);
        //// Check the parsed version.
        //// Reply with the handshake, or verify handshake.
        // Connect the session.
        const session = this.onConnect(conn, parsed.from);
        sessionCb?.(session);
      } catch (e) {
        // This is not a handshake message! This means that the workspace
        // connected to an older pid2 version, or viceversa, probably in a multiplayer
        // environment. We'll fall back to "v0" behavior.
        // TODO: Remove once we make sure this is not happening anywhere.
        //// parse data with `parseMsg`, do the session dance, and eventually call
        //// the callback and `this.handleMsg(parsed);`. Finally, fall through.
      }
      // Replace the listener with the normal message handler.
      conn.removeDataListener(firstMessageListener);
      conn.addDataListener((data) => this.handleMsg(this.parseMsg(data)));
    };
    conn.addDataListener(firstMessageListener);
  }
```

lhchavez (commented):
left a big comment that combines semantic neatness and also addresses pid2 version drift for the first release of this.

transport/impls/stdio/server.ts
jackyzha0: i'm hesitant about switching out the data listener on the fly like that because something might get dropped 😬 

in any case, i realized that the handshake bit was unused anyways so I just removed it

transport/transport.ts
```diff
@@ -515,3 +476,233 @@ export abstract class Transport<ConnType extends Connection> {
...
+      if (firstMessage) {
+        if (
+          !Value.Check(ControlMessageHandshakeResponseSchema, parsed.payload)
+        ) {
+          log?.warn(
+            `${
+              this.clientId
+            } -- received invalid handshake resp: ${JSON.stringify(parsed)}`,
+          );
+          return;
```
lhchavez, line 597: should we really `return` here? what happens in the scenario i outlined in the main PR? (multiplayers, one workspace is older than the other, both connecting to a newer or older pid2)

transport/impls/stdio/server.ts
lhchavez: hmm how would something get dropped? doesn't JS guarantee that as long as there are no await points, everything happens in the same threadloop? i.e. as long as you call `remove`+`add` in sequence, nothing _should_ be lost.

transport/transport.ts
jackyzha0: hm I can't find the comment in the main PR you are referring to, can you reiterate?

transport/impls/stdio/server.ts
jackyzha0: ah ok in my head there was a world in which

1. remove data listener
2. i/o event 
3. add data listener

could happen but it seems like this isn't the case as the event loop would only run after any synchronous code is done executing (i think?)

transport/transport.ts
jackyzha0: imo we shouldn't worry about mismatched river versions until after this PR lands

- not receiving a handshake message _should_ be treated as fatal/refuse the connection and will help us find mistakes faster
- pid2 flag audience is small enough (3 people) that just refreshing client for now is fine
- for ai chat, we can bake that compatability into the python river server

transport/impls/stdio/server.ts
lhchavez: i/o event can't happen between 1 and 3: all i/o events are "queued up" and are only "attended to" synchronously during the event loop.

https://nodejs.org/en/learn/asynchronous-work/overview-of-blocking-vs-non-blocking is a good mental model, everything goes through libuv (there's a non-trivial amount of libuv code in bun too). even if the i/o event actually happened, every JS code _expects_ to be single-threaded, so if bun didn't play by these rules and squeezed some code between 1 and 3, it would break so many existing codebases.

transport/transport.ts
lhchavez: oh so essentially implement what i'm suggesting but _only_ in ai chat. that seems reasonable, although i'd like us to start to build the backwards-compatibility muscle before it starts becoming load-bearing.

jackyzha0: yeah that makes sense, i think once this change is in place i'd be more more comfortable as this makes backwards-compatability much easier and less brittle than what we have now 👍

transport/session.ts
```diff
@@ -25,6 +25,22 @@ export abstract class Connection {
    * @param msg The message that was received.
    */
   abstract addDataListener(cb: (msg: Uint8Array) => void): void;
+  abstract removeDataListener(cb: (msg: Uint8Array) => void): void;
```
masad-frost, line 28: I prefer the disposable pattern where we return a handle to unlisten, just feels more ergonomic, no strong opinion.

transport/transport.ts
```diff
@@ -340,7 +284,7 @@ export abstract class Transport<ConnType extends Connection> {
    * @param msg The message to parse.
    * @returns The parsed message, or null if the message is malformed or invalid.
    */
-  protected parseMsg(msg: Uint8Array): OpaqueTransportMessage | null {
+  parseMsg(msg: Uint8Array): OpaqueTransportMessage | null {
```
masad-frost, line 287: why unprotected?

transport/transport.ts
```diff
@@ -241,11 +169,28 @@ export abstract class Transport<ConnType extends Connection> {
   }
 
   /**
-   * The downstream implementation needs to call this when a new connection is established
+   * This is called immediately after a new connection is established and we
+   * may or may not know the identity of the connected client.
+   *
+   * Calls addDataListener, addCloseListener, and addErrorListener on the connection
+   * with the appropriate handlers.
```
masad-frost, line 176: I'd just say it "attaches the required listeners" or something, just to avoid implementation details being in comments and things drifting too much over time.

transport/transport.ts
```diff
@@ -368,12 +312,23 @@ export abstract class Transport<ConnType extends Connection> {
     };
   }
 
+  /**
+   * Abstract method that receives a message with boot sequence.
+   * @param conn The connection object.
+   * @param sessionCb The callback to call with the established session.
+   * @returns The callback to call when a message is received.
+   */
+  abstract receiveWithBootSequence(
```
masad-frost, line 321: doesn't feel like this should be part of the abstract class, nothing calls it in the abstract class or externally, an implementation detail of what to do with a connection

transport/transport.ts
```diff
@@ -515,3 +476,220 @@ export abstract class Transport<ConnType extends Connection> {
...
+    );
+
+    conn.addCloseListener(() => {
+      if (!session) return;
+      log?.info(
+        `${this.clientId} -- connection (id: ${
+          conn.debugId
+        }) to ${client()} disconnected`,
+      );
+      this.onDisconnect(conn, session?.connectedTo);
```
masad-frost, line 640:
```suggestion
      this.onDisconnect(conn, session.connectedTo);
```

already checked if session is truthy

transport/transport.ts
```diff
@@ -515,3 +476,220 @@ export abstract class Transport<ConnType extends Connection> {
...
+            reason: 'VERSION_MISMATCH',
+          },
+        } satisfies Static<typeof ControlMessageHandshakeResponseSchema>);
+        conn.send(this.codec.toBuffer(responseMsg));
+        log?.warn(
+          `${this.clientId} -- received invalid handshake msg: ${JSON.stringify(
+            parsed,
+          )}`,
+        );
+        return;
```
masad-frost, line 676: hmm, seems like we have a ton of clean up here that we're not doing? Should we be terminating the connection after we send?

masad-frost: or is the idea that there's a recovery path here?

transport/transport.ts
```diff
@@ -515,3 +476,220 @@ export abstract class Transport<ConnType extends Connection> {
...
+      );
+    });
+  }
+
+  receiveWithBootSequence(
+    conn: ConnType,
+    sessionCb?: (sess: Session<ConnType>) => void,
+  ) {
+    const bootHandler = (data: Uint8Array) => {
+      const parsed = this.parseMsg(data);
```
masad-frost, line 658: I'm still very uncomfortable that failing to read/parse a message doesn't just break everything and silently fails.

transport/transport.ts
```diff
@@ -515,3 +476,220 @@ export abstract class Transport<ConnType extends Connection> {
...
+
+      const responseMsg = reply(parsed, {
+        type: 'HANDSHAKE_RESP',
+        status: {
+          ok: true,
+        },
+      } satisfies Static<typeof ControlMessageHandshakeResponseSchema>);
+      conn.send(this.codec.toBuffer(responseMsg));
+
+      const session = this.onConnect(conn, parsed.from);
```
masad-frost, line 687: feels like there needs to be more work done here to validate that the session is still valid for reconnects, e.g. checking sequence numbers as part of the payload or something, not sure.

transport/session.ts
```diff
@@ -25,6 +25,22 @@ export abstract class Connection {
...
+   * Handle adding a callback for when the connection is closed.
+   * This should also be called if an error happens.
+   * @param cb The callback to call when the connection is closed.
+   */
+  abstract addCloseListener(cb: () => void): void;
+
+  /**
+   * Handle adding a callback for when an error is received.
+   * This should only be used for logging errors, all cleanup
+   * should be delegated to addCloseListener.
```
masad-frost, line 40: :+1:

transport/impls/uds/connection.ts
```diff
@@ -21,6 +21,18 @@ export class UdsConnection extends Connection {
...
+  removeDataListener(cb: (msg: Uint8Array) => void): void {
+    this.input.off('data', cb);
+  }
+
+  addCloseListener(cb: () => void): void {
+    this.sock.on('close', cb);
+  }
+
+  addErrorListener(cb: (err: Error) => void): void {
+    this.sock.on('error', cb);
```
masad-frost, line 33: image

https://nodejs.org/api/net.html#event-error

Maybe we need to call sock.close here?

transport/session.ts
masad-frost: Might be worth being more explicit that errors that implementer needs to take care of "triggering a close event"

transport/transport.ts
```diff
@@ -515,3 +476,220 @@ export abstract class Transport<ConnType extends Connection> {
...
+      sessionCb?.(session);
+      conn.removeDataListener(bootHandler);
+      conn.addDataListener((data) => this.handleMsg(this.parseMsg(data)));
```
masad-frost, line 690: this pattern feels a little uncomfortable to read, i wonder if we can make it better.

Maybe simply inlining things in handleConnection would be eliminate the sessionCb and fix it

transport/transport.ts
```diff
@@ -515,3 +476,220 @@ export abstract class Transport<ConnType extends Connection> {
...
+        protocolVersion: PROTOCOL_VERSION,
+      } satisfies Static<typeof ControlMessageHandshakeRequestSchema>);
+      conn.send(this.codec.toBuffer(responseMsg));
+    } catch (err: unknown) {
+      // retry on failure
+      this.inflightConnectionPromises.delete(to);
+      if (attempt >= this.options.retryAttemptsMax) {
+        throw new Error(
+          `${this.clientId} -- connection to ${to} failed after ${attempt} attempts (${err}), giving up`,
+        );
```
masad-frost, line 565: Should probably log this. Also serializing `err` here might be painful to read or get good information from.

transport/transport.ts
```diff
@@ -515,3 +476,220 @@ export abstract class Transport<ConnType extends Connection> {
...
+      } else {
+        // exponential backoff + jitter
+        const jitter = Math.floor(Math.random() * this.options.retryJitterMs);
+        const backoffMs = this.options.retryIntervalMs * 2 ** attempt + jitter;
+        log?.warn(
+          `${this.clientId} -- connection to ${to} failed (${err}), trying again in ${backoffMs}ms`,
+        );
+        setTimeout(() => this.connect(to, attempt + 1), backoffMs);
```
masad-frost, line 573: any checks we need to do before blindly calling `this.connect`?

masad-frost (approved):
Looks good. Left some comments.

One meta comment is that I don't feel like I have a good grasp on how we deal with errors, logging is fine for observability, but it doesn't allow the consumer to handle things or at least receive a signal that things are absolutely broken. Maybe downstream from that, still feels like we can be asserting more and be less forgiving with undefined states.

transport/transport.ts
masad-frost: (or makes sense to override by a specific transport)

transport/transport.ts
jackyzha0: hm yeah idea is that client might send the proper message at some point in the future, though maybe that's bad anyways and we should just close the conn?

transport/transport.ts
jackyzha0: hm do you think this should be fatal?

transport/transport.ts
jackyzha0: yeah we don't do that in this PR, but https://github.com/replit/river/pull/55 adds the `instanceId` to the protocol handshake phase to do that check

transport/impls/uds/connection.ts
jackyzha0: I believe that's only for the actual `net.Server` itself, sock is `net.Socket` so it will call both `close` and `error`

transport/transport.ts
jackyzha0: nope, `this.connect` maintains its own invariants!

transport/transport.ts
masad-frost: Yeah probably, I mean if we get a bad packet, we'll probably continue getting bad ones, right? Like you can't recover that packet and we'll likely continue to receive bad packets

jackyzha0: gotcha, i'll do that in a later PR after we merge this stack

PR #54, opened 2024-02-24
[protocol] refactor acks

jackyzha0:
## Why
Because we don’t actually ack acks and we don’t re-transmit acks, sometimes an ack will be dropped during reconnection and we’ll end up with a message stuck in the sendBuffer forever :(

Also, we ack on every single message, it can significantly reduce our throughput we have high RPS (~10k/s)

## What
We can take inspiration from TCP here :) Instead of ack-ing based on `MsgId`, let’s just throw that out the window entirely and use sequence numbers. We also don’t need both `sendBuff` and `sendQueue`, we’ll just use `sendBuff`.

- Each transport message will have `seq` and `ack` fields and the `id` field will purely be for debug purposes.
- Each session will also track `seq` and `ack`
  - Ack: how many messages we’ve sent in the session
  - Seq: how many (unique) messages we’ve received in the session
- It will also have a grace expiry timeout which serves as our ping-pong strategy
  - This gets set on a msg send
  - This gets cleared when receiving a message
  - When this expires, we consider the connection dead and kill the underlying connection and try reconnecting
- We heartbeat on some interval to just send an empty message with the ack heartbeat bit set (this will have an empty payload)
- As ack is also included in each message, each outgoing message also acts like a micro-heartbeat
- On reconnect, send everything in our sendBuff
- If we receive a message with out-of-order seq, signal a session reconnect

### DIFF
--- a/router/client.ts
+++ b/router/client.ts
@@ -14,10 +14,9 @@ import { Server } from './server';
 import {
   OpaqueTransportMessage,
   ControlFlags,
-  msg,
   TransportClientId,
   isStreamClose,
-  closeStream,
+  PartialTransportMessage,
 } from '../transport/message';
 import { Static } from '@sinclair/typebox';
 import { nanoid } from 'nanoid';
@@ -192,7 +191,7 @@ export const createClient = <Srv extends Server<ServiceDefs>>(
       return handleRpc(
         transport,
         serverId,
-        input as object,
+        input as Record<string, unknown>,
         serviceName,
         procName,
       );
@@ -200,7 +199,7 @@ export const createClient = <Srv extends Server<ServiceDefs>>(
       return handleStream(
         transport,
         serverId,
-        input as object | undefined,
+        input as Record<string, unknown> | undefined,
         serviceName,
         procName,
       );
@@ -208,7 +207,7 @@ export const createClient = <Srv extends Server<ServiceDefs>>(
       return handleSubscribe(
         transport,
         serverId,
-        input as object,
+        input as Record<string, unknown>,
         serviceName,
         procName,
       );
@@ -216,7 +215,7 @@ export const createClient = <Srv extends Server<ServiceDefs>>(
       return handleUpload(
         transport,
         serverId,
-        input as object | undefined,
+        input as Record<string, unknown> | undefined,
         serviceName,
         procName,
       );
@@ -230,7 +229,7 @@ function createSessionDisconnectHandler(
   cb: () => void,
 ) {
   return (evt: EventMap['sessionStatus']) => {
-    if (evt.status === 'disconnect' && evt.session.connectedTo === from) {
+    if (evt.status === 'disconnect' && evt.session.to === from) {
       cb();
     }
   };
@@ -239,23 +238,18 @@ function createSessionDisconnectHandler(
 function handleRpc(
   transport: Transport<Connection>,
   serverId: TransportClientId,
-  input: object,
+  input: Record<string, unknown>,
   serviceName: string,
-  procName: string,
+  procedureName: string,
 ) {
   const streamId = nanoid();
-  const m = msg(
-    transport.clientId,
-    serverId,
+  transport.send(serverId, {
     streamId,
-    input,
     serviceName,
-    procName,
-  );
-
-  // rpc is a stream open + close
-  m.controlFlags |= ControlFlags.StreamOpenBit | ControlFlags.StreamClosedBit;
-  transport.send(m);
+    procedureName,
+    payload: input,
+    controlFlags: ControlFlags.StreamOpenBit | ControlFlags.StreamClosedBit,
+  });
 
   const responsePromise = new Promise((resolve) => {
     // on disconnect, set a timer to return an error
@@ -300,9 +294,9 @@ function handleRpc(
 function handleStream(
   transport: Transport<Connection>,
   serverId: TransportClientId,
-  init: object | undefined,
+  init: Record<string, unknown> | undefined,
   serviceName: string,
-  procName: string,
+  procedureName: string,
 ) {
   const streamId = nanoid();
   const inputStream = pushable({ objectMode: true });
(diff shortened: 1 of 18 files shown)


router/client.ts
```diff
@@ -192,31 +191,31 @@ export const createClient = <Srv extends Server<ServiceDefs>>(
       return handleRpc(
         transport,
         serverId,
-        input as object,
+        input as Record<string, unknown>,
```
lhchavez, line 194: :massage_man:

router/client.ts
```diff
@@ -458,49 +450,50 @@ function handleSubscribe(
 function handleUpload(
   transport: Transport<Connection>,
   serverId: TransportClientId,
-  input: object | undefined,
```
lhchavez, line 461: oops!

transport/impls/stdio/connection.ts
```diff
@@ -21,7 +21,8 @@ export class StreamConnection extends Connection {
   }
 
   send(payload: Uint8Array) {
-    return this.output.write(MessageFramer.write(payload));
+    setImmediate(() => this.output.write(MessageFramer.write(payload)));
```
lhchavez, line 24: :thinking: https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate

> Deprecated: This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible; see the [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate#browser_compatibility) at the bottom of this page to guide your decision. Be aware that this feature may cease to work at any time.

lhchavez: ah nvm, this is only going to be run in a bun context. in any case, it might be good mentioning why this is run as an immediate vs. doing it in the same loop iteration.

transport/session.ts
```diff
@@ -75,63 +85,218 @@ export const DISCONNECT_GRACE_MS = 3_000; // 3s
...
+
+    // setup heartbeat
+    this.heartbeatMisses = 0;
+    this.heartbeat = setInterval(
+      () => this.sendHeartbeat(),
+      HEARTBEAT_INTERVAL_MS,
+    );
+  }
+
+  send(msg: PartialTransportMessage, skipRetry?: boolean): string {
```
lhchavez, line 153: might be worth adding a lil docstring and explain what `skipRetry` implies.

also, if it's not added to the send buffer, what does that do for reconnections?

transport/session.ts
```diff
@@ -75,63 +85,218 @@ export const DISCONNECT_GRACE_MS = 3_000; // 3s
...
+
+  replaceWithNewConnection(newConn: ConnType) {
+    this.closeStaleConnection(this.connection);
+    this.cancelGrace();
+    this.connection = newConn;
   }
 
   beginGrace(cb: () => void) {
-    this.graceExpiryTimeout = setTimeout(() => {
+    this.disconnectionGrace = setTimeout(() => {
```
lhchavez, line 254: https://www.youtube.com/watch?v=cOTmvgbkOsc

dis-cooooooo-neeec-tion graaaaace ~

lhchavez (commented):
first round done! will finish tomorrow too

transport/session.ts
```diff
@@ -75,63 +85,218 @@ export const DISCONNECT_GRACE_MS = 3_000; // 3s
...
+  }
+
+  sendBufferedMessages() {
+    if (!this.connection) {
+      const msg = `${this.from} -- tried sending buffered messages without a connection (if you hit this code path something is seriously wrong)`;
+      log?.error(msg);
+      throw new Error(msg);
+    }
+
+    for (const msg of this.sendBuffer) {
```
lhchavez, line 212: clarifying, we don't yeet the contents of `sendBuffer` because we might be in a situation where a client is sent the buffer, but does not ack, so after _another_ reconnection we send it all back, right?

jackyzha0: yup, this + the fact that we didn't track acks before was actually the root cause of us sometimes dropping acks

transport/session.ts
jackyzha0: this is just a minor optimization, ack messages themselves are just 'empty' messages that carry the seq/ack information so we don't want them to build up while we are reconnecting (imagine a case where we disconnect for ~10h and get like 10k acks in the queue, it would be silly to blast them back to the other peer when we reconnect as the real messages we will have queued up will carry the same info)

jackyzha0: i'll add a docstring here to explain that

transport/impls/stdio/connection.ts
```diff
@@ -31,7 +31,16 @@ export class StreamConnection extends Connection {
...
-    return this.output.write(MessageFramer.write(payload));
+    // many of our tests are structured in a way that is roughly
+    // 1. const msg = {...}
+    // 2. const msgId = clientTransport.send(serverTransport.clientId, msg)
+    // 3. await waitForMessage(serverTransport, (msg) => msg.id === msgId)
+    //
+    // however, because we pipe two streams into each other, the stream transport
+    // actually receives the message before we even attach the listener in L3.
+    // we use setImmediate() here to schedule it for the next check phase of the event loop
+    setImmediate(() => this.output.write(MessageFramer.write(payload)));
```
lhchavez, line 42: hmm wait, there could be some problems here: do we have a way of ensuring that _nothing_ calls `this.output.write` in a blocking context? otherwise we could end up in a situation where these writes become out of order.

also, we could also have some fun races with `this.send()` and `this.close()` where the socket is closed before the output is sent!

jackyzha0: aren't `setImmediate` callbacks in FIFO order? the writes themselves should always be in order (this is the only place that writes to `this.output`)

ah yeah good point, lemme refactor to account for the race

jackyzha0: ended up just refactoring the test so we don't have this pattern anymore

lhchavez (approved):
:sheep:

PR #66, opened 2024-03-28
[WIP] Implement authorize callback for handshakes

Monkatraz:
Very WIP.

Ideally, we could make authorization simple and just do it all in the handshake. I suspect the way this PR does it is not nearly enough though. Likely the server's authorize callback will need async support, and I suspect clients will need a way to handle their authorization automatically, probably through a callback that gets them their authorization details whenever they want.

Tons of open questions here:
- how should we handle transparent reconnects? are you allowed to reuse your authorization?
- how do we prevent someone reading pid2 memory and spoofing another client via their authorization (which may be in memory)? is this even a concern? 
- should we allow some (or even all) messages to be signed, especially ones which may cause destructive actions? also, again, is this even a concern?
- what's the best way to expose sessions and their details to procedures?
- how do we add metadata to this, so that we can attach information like username and user id?
- do we need middleware so that we can gate procedures based on permissions/role? are we ever going to need a permissions system?

Other comments:
- seems like [tRPC recommends authorizing via middleware or in resolvers](https://trpc.io/docs/server/authorization)

### DIFF
--- a/__tests__/fixtures/matrix.ts
+++ b/__tests__/fixtures/matrix.ts
@@ -16,6 +16,7 @@ interface TestMatrixEntry {
       simulatePhantomDisconnect: () => void;
       getClientTransport: (
         id: TransportClientId,
+        authorization?: unknown,
       ) => ClientTransport<Connection>;
       getServerTransport: () => ServerTransport<Connection>;
       restartServer: () => Promise<void>;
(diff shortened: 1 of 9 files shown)


jackyzha0:
- i do think that _some_ sort of middleware system will be useful at some point (esp as we think about adding tracing too)
- procedures can already access session via `ctx.session` i would imagine we could put custom handshake info there too
- i wonder if theres a way we can include the handshake info as a typebox type and have the protocol be parametric over the handshake info
- hm about security, we probably wouldn't need to store incoming authorization right? auth should probably be done per session (basically anytime we handshake)

PR #67, opened 2024-03-29
client-transport: leaky bucket connection retry rate limit

jackyzha0:
## why
- our retry logic didn't handle cases where a connection appears successful (i.e. we get an 'open' message) and then is immediately closed
  - this can end up in connection spam
- we should globally add some form of rate-limiting at a transport-level per node it is attempting to connect to

## what changed
- add leaky bucket rate limiting so we dont spam retry even on connections that open and then immediately close
- moved all retry args to `connectionRetryOptions`
    - add arg `budgetRestoreIntervalMs` that controls how often we refill the bucket
- `connect` promise lifetime is now accurate across retries
- add test `repeated connections that close instantly causes retry limit to still be reached`
- Split `TransportOptions` to `ServerTransportOptions` and `ClientTransportOptions`. We now only export the shape that is accepted by the transport constructor, instead of the options internal to the transports.
- Isolate default values in tests to use their own instead of relying on global defaults.

### DIFF
--- a/__tests__/negative.test.ts
+++ b/__tests__/negative.test.ts
@@ -14,7 +14,6 @@ import { NaiveJsonCodec } from '../codec';
 import { Static } from '@sinclair/typebox';
 import { WebSocketClientTransport } from '../transport/impls/ws/client';
 import { ProtocolError } from '../transport/events';
-import { defaultTransportOptions } from '../transport/transport';
 import WebSocket from 'ws';
 
 describe('should handle incompatabilities', async () => {
@@ -98,11 +97,18 @@ describe('should handle incompatabilities', async () => {
       ws.close();
     };
 
+    const maxAttempts = 10;
     wss.on('connection', serverWsConnHandler);
-    const clientTransport = new WebSocketClientTransport(() => {
-      const ws = createLocalWebSocketClient(port);
-      return Promise.resolve(ws);
-    }, 'client');
+    const clientTransport = new WebSocketClientTransport(
+      () => {
+        const ws = createLocalWebSocketClient(port);
+        return Promise.resolve(ws);
+      },
+      'client',
+      {
+        connectionRetryOptions: { maxAttempts },
+      },
+    );
     clientTransport.tryReconnecting = false;
 
     const errMock = vi.fn();
@@ -113,17 +119,11 @@ describe('should handle incompatabilities', async () => {
       clientTransport.close();
     });
 
-    for (
-      let i = 0;
-      i < defaultTransportOptions.maxReconnectionBurstAttempts;
-      i++
-    ) {
+    for (let i = 0; i < maxAttempts; i++) {
       void clientTransport.connect('SERVER');
     }
 
-    expect(conns).toBeLessThan(
-      defaultTransportOptions.maxReconnectionBurstAttempts,
-    );
+    expect(conns).toBeLessThan(maxAttempts);
   });
 
   test('incorrect client handshake', async () => {
--- a/__tests__/negative.test.ts
+++ b/__tests__/negative.test.ts
@@ -1,12 +1,4 @@
-import {
-  afterAll,
-  assert,
-  describe,
-  expect,
-  onTestFinished,
-  test,
-  vi,
-} from 'vitest';
+import { afterAll, describe, expect, onTestFinished, test, vi } from 'vitest';
 import http from 'node:http';
 import { testFinishesCleanly, waitFor } from './fixtures/cleanup';
 import {
@@ -23,7 +15,6 @@ import { Static } from '@sinclair/typebox';
 import { WebSocketClientTransport } from '../transport/impls/ws/client';
 import { ProtocolError } from '../transport/events';
 import { defaultTransportOptions } from '../transport/transport';
-import { bindLogger } from '../logging';
 import WebSocket from 'ws';
 
 describe('should handle incompatabilities', async () => {
--- a/__tests__/negative.test.ts
+++ b/__tests__/negative.test.ts
@@ -1,4 +1,12 @@
-import { afterAll, describe, expect, onTestFinished, test, vi } from 'vitest';
+import {
+  afterAll,
+  assert,
+  describe,
+  expect,
+  onTestFinished,
+  test,
+  vi,
+} from 'vitest';
 import http from 'node:http';
 import { testFinishesCleanly, waitFor } from './fixtures/cleanup';
 import {
@@ -14,6 +22,9 @@ import { NaiveJsonCodec } from '../codec';
 import { Static } from '@sinclair/typebox';
 import { WebSocketClientTransport } from '../transport/impls/ws/client';
 import { ProtocolError } from '../transport/events';
+import { defaultTransportOptions } from '../transport/transport';
+import { bindLogger } from '../logging';
+import WebSocket from 'ws';
 
 describe('should handle incompatabilities', async () => {
   const server = http.createServer();
@@ -34,7 +45,7 @@ describe('should handle incompatabilities', async () => {
     await clientTransport.connect(serverTransport.clientId);
     await waitFor(() => expect(serverTransport.connections.size).toBe(1));
 
-    const errMock = vi.fn<[], unknown>();
+    const errMock = vi.fn();
     clientTransport.addEventListener('protocolError', errMock);
     onTestFinished(async () => {
       clientTransport.removeEventListener('protocolError', errMock);
@@ -58,12 +69,13 @@ describe('should handle incompatabilities', async () => {
   });
 
   test('conn failure, retry limit reached', async () => {
+    vi.useFakeTimers({ shouldAdvanceTime: true });
     const clientTransport = new WebSocketClientTransport(
       () => Promise.reject(new Error('fake connection failure')),
       'client',
     );
     const serverTransport = new WebSocketServerTransport(wss, 'SERVER');
-    const errMock = vi.fn<[], unknown>();
+    const errMock = vi.fn();
     clientTransport.addEventListener('protocolError', errMock);
     onTestFinished(async () => {
       clientTransport.removeEventListener('protocolError', errMock);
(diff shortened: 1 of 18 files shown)


__tests__/negative.test.ts
```diff
@@ -77,19 +80,57 @@ describe('should handle incompatabilities', async () => {
...
-    expect(errMock).toHaveBeenCalledTimes(1);
+    await waitFor(() => expect(errMock).toHaveBeenCalledTimes(1));
     expect(errMock).toHaveBeenCalledWith(
       expect.objectContaining({
         type: ProtocolError.RetriesExceeded,
       }),
     );
   });
 
+  test('repeated connections that close instantly causes retry limit to still be reached', async () => {
```
masad-frost, line 96: :man_cook:

__tests__/negative.test.ts
```diff
@@ -77,19 +80,57 @@ describe('should handle incompatabilities', async () => {
...
+      return Promise.resolve(ws);
+    }, 'client');
+    clientTransport.tryReconnecting = false;
+
+    const errMock = vi.fn();
+    clientTransport.addEventListener('protocolError', errMock);
+    onTestFinished(() => {
+      wss.off('connection', immediatelyCloseWebSocket);
+      clientTransport.removeEventListener('protocolError', errMock);
+    });
```
masad-frost, line 115: when do you use this vs just doing it at the end?

transport/transport.ts
```diff
@@ -460,13 +462,46 @@ export abstract class Transport<ConnType extends Connection> {
   }
 }
 
+class LeakyBucketLimit {
```
masad-frost, line 465: might be worth a test on this class?

masad-frost: wonder if building the backoff into this makes sense, like make it return a backoff exponent instead of just a limit? We can get rid of the attempt counter altogether that way and maintain backoffs across connections, instead of just per connection cycle.

transport/transport.ts
```diff
@@ -637,9 +702,9 @@ export abstract class ClientTransport<
     conn.send(this.codec.toBuffer(requestMsg));
   }
 
-  protected onDisconnect(conn: ConnType, session: Session<ConnType>): void {
-    this.inflightConnectionPromises.delete(session.to);
```
masad-frost, line 641: why is this not needed anymore?

transport/transport.ts
```diff
@@ -581,21 +622,46 @@ export abstract class ClientTransport<
...
     let reconnectPromise = this.inflightConnectionPromises.get(to);
     if (!reconnectPromise) {
-      reconnectPromise = this.createNewOutgoingConnection(to).then((conn) => {
-        // only send handshake once per attempt
-        this.sendHandshake(to, conn);
-        return conn;
-      });
+      // check budget
+      const budgetConsumed = this.retryBudget.getBudgetConsumed(to);
+      if (budgetConsumed >= this.options.retryAttemptsMax) {
```
masad-frost, line 637: I think this option might be a misnomer now.

transport/transport.ts
```diff
@@ -61,6 +62,7 @@ export const defaultTransportOptions: TransportOptions = {
   retryIntervalMs: RECONNECT_INTERVAL_MS,
   retryJitterMs: RECONNECT_JITTER_MAX_MS,
   retryAttemptsMax: 5,
+  retryBudgetRestoreRateMs: 500,
```
masad-frost, line 65: Feels like way too low, I think on the order of seconds makes more sense to me.

masad-frost (commented):
Nice quick turn around.

Left some comments, lemme know what you think.

transport/transport.ts
jackyzha0: we moved this to be done after a connection fails, `onDisconnect` is only called for connections that have completed a handshake so we need to also handle the case when the server just received a connection and it closed immediately without handshake completing

transport/transport.ts
jackyzha0: oh true 🤔 lemme try my hand at a refactor

__tests__/negative.test.ts
```diff
@@ -77,19 +80,57 @@ describe('should handle incompatabilities', async () => {
...
       }),
     );
   });
 
+  test('repeated connections that close instantly causes retry limit to still be reached', async () => {
+    const onConnMock = vi.fn();
+    const immediatelyCloseWebSocket = (ws: WebSocket) => {
+      onConnMock();
+      ws.close();
+    };
```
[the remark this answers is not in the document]
jackyzha0: crazyyy

__tests__/negative.test.ts
jackyzha0: this handles cases where a test fails and throws before we reach the end of the test (its kinda like a .finally)

transport/transport.ts
masad-frost: slack'd you a suggestion

transport/transport.ts
```diff
@@ -41,26 +42,17 @@ import { coerceErrorString } from '../util/stringify';
 export type TransportStatus = 'open' | 'closed' | 'destroyed';
 
 export type TransportOptions = {
-  retryIntervalMs: number;
+  maxReconnectionBurstAttempts: number;
```
masad-frost, line 45: good name

transport/transport.ts
```diff
@@ -41,26 +42,17 @@ import { coerceErrorString } from '../util/stringify';
...
+  retryBudgetRestoreIntervalMs: number;
+  retryBaseIntervalMs: number;
   retryJitterMs: number;
```
masad-frost, line 48: this should be `maxRetryJitterMs` right?

masad-frost, line 48: maybe we should put these under a single object, `retryOptions` or something.

transport/rateLimit.ts
```diff
@@ -0,0 +1,56 @@
...
+    return this.options.leakIntervalMs * this.options.maxBurst;
+  }
+
+  getBackoffMs(user: TransportClientId) {
+    if (!this.budgetConsumed.has(user)) return 0;
+
+    const exponent = Math.max(0, this.getBudgetConsumed(user) - 1);
+    const jitter = Math.floor(Math.random() * this.options.jitterMs);
+    const backoffMs = this.options.baseIntervalMs * 2 ** exponent + jitter;
+    return backoffMs;
```
masad-frost, line 42: We should probably have a maximum backoff now that we don't have max attempts and things are time based.

transport/rateLimit.ts
```diff
@@ -0,0 +1,56 @@
...
+    this.intervalHandle = setInterval(() => {
+      for (const [user, budgetConsumed] of this.budgetConsumed.entries()) {
+        const newBudget = budgetConsumed - 1;
+        if (newBudget === 0) {
+          this.budgetConsumed.delete(user);
+        } else {
+          this.budgetConsumed.set(user, newBudget);
+        }
+      }
+    }, this.options.leakIntervalMs);
```
masad-frost, line 29: I really dislike that budgets start getting refreshed as soon as you consume. If your `leakIntervalMs` isn't much much higher than the duration it takes to establish a connection, then you're probably not really implementing exponential backoff, you just have normal backoff. You can also end up in a situation where you have 2 tries 1ms apart that get 0 backoff because of the timing (e.g. right before the interval hits).

transport/rateLimit.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    const exponent = Math.max(0, this.getBudgetConsumed(user) - 1);
+    const jitter = Math.floor(Math.random() * this.options.jitterMs);
+    const backoffMs = this.options.baseIntervalMs * 2 ** exponent + jitter;
+    return backoffMs;
+  }
+
+  consumeBudget(user: TransportClientId) {
+    this.budgetConsumed.set(user, this.getBudgetConsumed(user) + 1);
+  }
```
masad-frost, line 47: should these be the one function?

masad-frost (changes requested):
I think we're losing our backoff with this unfortunately. We can make the drainage time astronomical which would put us at risk of not being able to reconnect when we should be able to. My recommendation is  to defer it until we connect, the example i sent you does that. Otherwise we're running the risk of little to no backoff.

I just tried it in the console, I can barely get the backoff to grow unless i'm spamming consume budget, feel free to try it.

```js
class LeakyBucketRateLimit {
    constructor(options) {
        this.options = options;
        this.budgetConsumed = new Map();
        // start leaking
        this.intervalHandle = setInterval(() => {
            for (const [user, budgetConsumed] of this.budgetConsumed.entries()) {
                const newBudget = budgetConsumed - 1;
                console.log('lower budget for', user, budgetConsumed)
                if (newBudget === 0) {
                    this.budgetConsumed.delete(user);
                }
                else {
                    this.budgetConsumed.set(user, newBudget);
                }
            }
        }, this.options.leakIntervalMs);
    }
    get drainageTimeMs() {
        return this.options.leakIntervalMs * this.options.maxBurst;
    }
    getBackoffMs(user) {
        if (!this.budgetConsumed.has(user))
            return 0;
        const exponent = Math.max(0, this.getBudgetConsumed(user) - 1);
        const jitter = Math.floor(Math.random() * this.options.jitterMs);
        const backoffMs = this.options.baseIntervalMs * 2 ** exponent + jitter;
        return backoffMs;
    }
    consumeBudget(user) {
        this.budgetConsumed.set(user, this.getBudgetConsumed(user) + 1);
    }
    getBudgetConsumed(user) {
        return this.budgetConsumed.get(user) ?? 0;
    }
    close() {
        clearInterval(this.intervalHandle);
    }
}
l = new LeakyBucketRateLimit({maxBurst: 1000, leakIntervalMs: 2000, baseIntervalMs: 250, jitterMs: 500 })
u = 'user'
```

and then try this in the console.

```js
l.consumeBudget(u)
console.log('backoff', l.getBackoffMs(u))
console.log('consumed', l.getBudgetConsumed(u))
```
it doesn't feel right to me. WDYT?

transport/rateLimit.ts
jackyzha0: do you think we should start a separate interval for each budget?

masad-frost: Yeah let's do that, probably enough.

masad-frost (approved):

transport/rateLimit.ts
```diff
@@ -1,10 +1,35 @@
...
-  jitterMs: number;
+  /**
+   * The maximum random jitter to add to the total backoff time.
+   */
+  maxJitterMs: number;
+  /**
+   * The maximum amount of time to wait before retrying a connection.
+   * This does not include the jitter
+   */
+  maxBackoffMs: number;
```
jackyzha0, line 22: should probably mention that we default to an exponential backoff and cap the result by this number

transport/rateLimit.ts
```diff
@@ -1,10 +1,35 @@
...
+  /**
+   * The maximum number of times to retry a connection before giving up.
+   * This persists across connections but starts reseting after every succesful
+   * connection, the restoration period depends on {@link Connection.retryBudgetRestoreIntervalMs}
+   */
+  maxAttempts: number;
```
jackyzha0, line 28: when i read this, it feels weird that we need to define this along with `maxBackoffMs`, imo should be one or the other? 

maybe `attemptCapacity` is better at conveying the semantics of how we actually use it to calculate

transport/transport.ts
```diff
@@ -630,6 +630,11 @@ export abstract class ClientTransport<
             throw new Error('transport state is no longer open');
           }
 
+          // After a successful connection, we start restoring the budget
+          // so that the next time we try to connect, we don't hit the client
+          // with backoff forever.
+          this.retryBudget.startRestoringBudget(to);
```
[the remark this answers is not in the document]
masad-frost: no u

transport/rateLimit.ts
```diff
@@ -43,14 +35,53 @@ export class LeakyBucketRateLimit {
   }
 
   consumeBudget(user: TransportClientId) {
+    // If we're consuming again, let's ensure that we're not leaking
+    this.stopLeak(user);
```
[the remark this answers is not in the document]
masad-frost: Intuitively, this feels right. idk

transport/rateLimit.ts
masad-frost: Yeah something feels off now that we're not leaking while we're attempting. Good name

transport/rateLimit.ts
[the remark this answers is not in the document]
jackyzha0: ya that makes sense too, just didnt get that immediately haha

jackyzha0:
image

me reviewing my own pr

PR #71, opened 2024-04-02
Separate services and procedures types, give each procedure form its own type

Monkatraz:
Before we change how you construct services, I wanted to change how the procedure types worked so that we could individually type each procedure form. That's what this PR is.

Procedure types are now in `router/procedures.ts`, and service types and the builder for services is now in `router/services.ts`. I gave each procedure type (upload, rpc, etc.) its own type, and then made `Procedure` use those types. I think this might technically break the TS stress test, but the actual constructed type works, it's just TS complains that it _may_ not work, which is strange.

### DIFF
--- a/__tests__/fixtures/services.ts
+++ b/__tests__/fixtures/services.ts
@@ -1,5 +1,5 @@
 import { Type } from '@sinclair/typebox';
-import { ServiceBuilder } from '../../router/builder';
+import { ServiceBuilder } from '../../router/services';
 import { Err, Ok } from '../../router/result';
 import { Observable } from './observable';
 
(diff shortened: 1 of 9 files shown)


Monkatraz:
ok yeah TS is unhappy, blegh. tempted to just modify the test so that it passes (remove the last three procedures) because we're going to be getting rid of this limitation anyways

bradymadden97:
I think we will want to build to the end state to know we can get there before removing the stress tests. This is clearly a tricky problem and I'm afraid we need to know a direction works completely before we commit to it, rather than introducing changes that require us to undo tests.

Monkatraz:
> I think we will want to build to the end state to know we can get there before removing the stress tests.

Oh yeah I meant to mention that it's probably best if we wait to merge until we have a whole stack going, whoops

Monkatraz:
Closing this PR and making the other one the main one

PR #76, opened 2024-04-03
`ServiceSchema` implementation

Monkatraz:
TypeScript has nearly taken my soul.

This PR replaces our current way of creating services. You now create them like this:
```ts
const MyServiceSchema = ServiceSchema.define(
  () => ({ count: 0 }),
  {
    increment: Procedure.rpc({
      input: Type.Object({ amount: Type.Number() }),
      output: Type.Object({ current: Type.Number() }),
      async handler(ctx, input) {
        ctx.state.count += input.amount;
        return Ok({ current: ctx.state.count });
      }
    }),
  },
);
```
or like this, if you have no state:
```ts
const MyServiceSchema = ServiceSchema.define({
  add: Procedure.rpc({
    input: Type.Object({ a: Type.Number(), b: Type.Number() }),
    output: Type.Object({ result: Type.Number() }),
    async handler(ctx, input) {
      return Ok({ result: input.a + input.b });
    }
  }),
});
```

Either way, you're given back a `ServiceSchema`, which you can basically treat like a class for your service. Services also no longer have names, instead you provide a dictionary to the server where the keys are the names, and the values are the service schemas:
```ts
const server = createServer(transport, { myService: MyServiceSchema })
```

This way of constructing services gets us 90% out of recursive TypeScript hell, and is also intended to make writing services easier, primarily through better legibility.

Some other changes:
- To be both more accurate and to avoid type issues related to `never`, procedures can now always emit `RiverUncaughtSchema` errors and it is now allowed to omit the `errors` property when creating procedures. If you don't provide the prop, it just defaults to `Type.Never()`.
- Procedures now must be constructed via the `Procedure` object, e.g. `Procedure.rpc(...)`. This prevents footguns that can happen by manually providing procedures to services (TS gets confused and doesn't type check your procedure!), and also lets us enable ergonomics (like omitting the `errors` prop). The way the footgun is forcibly avoided is purely through the type system - procedures created by these constructors are "branded" with an inaccessible unique symbol (which is not actually there), and `ServiceSchema.define` accepts _only_ branded procedures.


router/procedures.ts
```diff
@@ -0,0 +1,495 @@
...
+  init: Init;
+  input: I;
+  output: O;
+  errors: E;
+  handler: StreamProcedure<State, I, O, E, Init>['handler'];
+}): Branded<StreamProcedure<State, I, O, E, Init>>;
+
+// signature: no init with default errors
+function stream<State, I extends PayloadType, O extends PayloadType>(def: {
+  input: I;
```
jackyzha0, line 442: i noticed that when we dont have init we actually just omit it whereas with errors we mark it optional, is there a difference and why do we treat them differently?

router/result.ts
```diff
@@ -44,14 +46,14 @@ export type Result<T, E> =
       payload: E;
     };
 
-export function Ok<T, E>(payload: T): Result<T, E> {
+export function Ok<const T, const E>(payload: T): Result<T, E> {
```
jackyzha0, line 49: what do const generics do for us here?

router/server.ts
```diff
@@ -26,16 +31,15 @@ import {
...
 import { Connection } from '../transport';
 import { coerceErrorString } from '../util/stringify';
 
 /**
  * Represents a server with a set of services. Use {@link createServer} to create it.
  * @template Services - The type of services provided by the server.
  */
-export interface Server<Services> {
-  services: Services;
+export interface Server<Services extends ServiceSchemaMap> {
```
jackyzha0, line 41: yay nice

router/server.ts
```diff
@@ -422,7 +432,7 @@ class RiverServer<Services extends ServiceDefs> {
   private getContext(service: AnyService) {
     const context = this.contextMap.get(service);
     if (!context) {
-      const err = `${this.transport.clientId} -- no context found for ${service.name}`;
+      const err = `${this.transport.clientId} -- no context found for service`;
```
jackyzha0, line 435: hm any way we can get a better error here?

router/services.ts
```diff
@@ -0,0 +1,310 @@
...
+   *   },
+   * );
+   * ```
+   */
+  static define<
+    State extends object,
+    Procedures extends BrandedProcedureMap<State>,
+  >(
+    initState: () => State,
+    procedures: Procedures,
```
jackyzha0, line 219: from comment above: maybe we can turn this into `{ initState: () => State, procedures: Procedures }` just so reading the plaintext version of the code is easier

router/services.ts
```diff
@@ -0,0 +1,310 @@
...
+      return new ServiceSchema(() => stateOrProcedures(), procedures);
+    }
+
+    return new ServiceSchema(() => ({}), stateOrProcedures);
+  }
+
+  /**
+   * Serializes this schema's procedures into a plain object that is JSON compatible.
+   */
+  serialize(): object {
```
jackyzha0, line 271: maybe for another PR but i'd like to have a helper type that reconstructs `ServiceSchemaMap` from a bunch of serialized services

__tests__/fixtures/services.ts
```diff
@@ -10,210 +11,184 @@ export const EchoRequest = Type.Object({
...
-      async handler(_ctx, msgStream, returnStream) {
-        for await (const { ignore, msg, end } of msgStream) {
-          if (!ignore) {
-            returnStream.push(Ok({ response: msg }));
-          }
-
-          if (end) {
-            returnStream.end();
-          }
+export const TestServiceSchema = ServiceSchema.define(() => ({ count: 0 }), {
```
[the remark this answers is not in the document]
Monkatraz: The only thing calling `instantiate` on the schemas is the server, so it's never at the end when defining services. But yes, I agree with this critique - problem is I'm not sure if there is anything I can do about it, because the solution you offered (using an object) makes TypeScript fail to figure out what the state's type should be in the procedures. So we have to use params like this.

router/index.ts
```diff
@@ -1,17 +1,24 @@
...
   PayloadType,
-} from './builder';
-export { buildServiceDefs } from './defs';
-export type { ServiceDefs } from './defs';
+  ProcedureMap,
+  ProcedureResult,
+  RPCProcedure,
+  UploadProcedure,
+  SubscriptionProcedure,
+  StreamProcedure,
```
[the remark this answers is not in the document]
Monkatraz: If you're saying this thinking that these are branded types, they're not. `Procedure.rpc` returns a `Branded<RPCProcedure>`, rather than `RPCProcedure` itself already being branded.

router/result.ts
Monkatraz: If you have a strictly typed result like this:
```ts
type MyOutput = Type.Object({ str: Type.Literal('foo') })
type MyResult = Result<Static<typeof MyOutput>, never>
```
then, counter-intuitively, this may not compile (without const generics):
```ts
const myResult: MyResult = Ok({ str: 'foo' })
```
this is because the object inside the `Ok` is typed broadly, like `{ str: string }`. It will compile if you do this though:
```ts
const myResult: MyResult = Ok({ str: 'foo'} as const)
```
which is annoying to have to type every time. The solution is to use const generics, which just does the `as const` for you (literally, that's all it does).

router/services.ts
Monkatraz: TypeScript is mean and doesn't let us use an object :(

router/services.ts
Monkatraz: How would the state be serialized? Do we even care about state ever being serialized?

router/procedures.ts
```diff
@@ -0,0 +1,495 @@
...
+// signature: explicit errors
+function rpc<
+  State,
+  I extends PayloadType,
+  O extends PayloadType,
+  E extends RiverError,
+>(def: {
+  input: I;
+  output: O;
+  errors: E;
```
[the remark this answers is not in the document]
Monkatraz: It is. I'm not sure if that's entirely desirable, but I figured since technically the uncaught error type applies to any procedure it would be strange if you weren't allowed to return an uncaught error type.

router/procedures.ts
```diff
@@ -0,0 +1,495 @@
...
+  Init extends PayloadType
+    ? Ty extends 'upload' ? UploadProcedure<State, I, O, E, Init>
+    : Ty extends 'stream' ? StreamProcedure<State, I, O, E, Init>
+    : never
+  : Ty extends 'rpc' ? RPCProcedure<State, I, O, E>
+  : Ty extends 'upload' ? UploadProcedure<State, I, O, E>
+  : Ty extends 'subscription' ? SubscriptionProcedure<State, I, O, E>
+  : Ty extends 'stream' ? StreamProcedure<State, I, O, E>
+  : never
+);
```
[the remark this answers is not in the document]
Monkatraz: yeah sometimes you have to tell Prettier to go away to make readable types lol

router/procedures.ts
```diff
@@ -0,0 +1,495 @@
...
+    PayloadType,
+    PayloadType,
+    RiverError,
+    PayloadType | null
+  >['handler'];
+}) {
+  return init !== undefined && init !== null
+    ? { type: 'upload', init, input, output, errors, handler }
+    : { type: 'upload', input, output, errors, handler };
+}
```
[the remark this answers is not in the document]
Monkatraz: Not sure, regardless we should have one, will add

router/server.ts
Monkatraz: Probably have to pass in the service name, purely for debugging. Will add

__tests__/fixtures/services.ts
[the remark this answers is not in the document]
Monkatraz: Another thing about using a builder pattern or nested objects is that the indentation level gets kind of absurdly high. A nice thing about doing it this way is that you can do this:
```ts
function init() {
  return { count: 1 }
}

const MyService = ServiceSchema.define(init, {
  // now we're in the lowest indentation level we can reasonably get, only one indent
}
```
pretty much every other pattern adds two or even three indent units.

This is also why I really liked this class based syntax:
```ts
class MyState {
  count = 1
}

const MyService = ServiceSchema.define(MyState, {
  // ...
}
```
still mostly supported though, as you can just do this:
```ts
const MyService = ServiceSchema.define(() => new MyState(), {
  // ...
}
```

router/result.ts
Monkatraz: I think this wasn't needed before because TS was doing weird type coercion and honestly was probably doing something unsound

router/procedures.ts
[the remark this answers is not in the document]
jackyzha0: yup just making sure :)

router/result.ts
jackyzha0: ok cool :)

router/services.ts
jackyzha0: i think ok for state to not be serialized, i meant for the client type (the result of `Server<typeof serviceDefs>>` normally

mostly to help us ensure we have a way to go from JSON schema generated by a non-TS implementation of Quartz (e.g. Python or Protobufs ala AI chat) to something we can consume in TS

router/services.ts
Monkatraz: I wonder if we should just totally separate procedure definition from state stuff... maybe something like:
```ts
const MyServiceScaffold = ServiceSchema.scaffold({
  // could allow this to be async too?
  initializeState: () => ({ count: 1 }),
  setup: async () => { /* ... */ },
  teardown: async () => { /* ... */ },
})

// ProcedureMap<StateOf<MyServiceScaffold>>
const someProcedures = MyServiceScaffold.procedures({
  // ...
})
// alternatively, maybe calling `procedures` mutates the scaffold?

const MyService = MyServiceScaffold.finalize({
  ...someProcedures,
  // ...
})
```
this would let you define a service over multiple files, kind of like "merging" multiple services. I know @bradymadden97 desired something like this, because of the sheer amount of procedures the LSP service may have to have

jackyzha0: oh i do like the ability to scaffold like that, i think procedures returning a new copy instead of mutating is good

let's do this! we can leave the setup/teardown out for now i can do it in a follow up

Monkatraz: Yeah and I suppose this lets you do the builder pattern if you want too:
```ts
const MyService = ServiceSchema
  .scaffold({
    initializeState: () => ({ count: 1 })
  })
  .finalize({
    foo: Procedure.rpc(/* ... */)
  })
```
I would like a better name than `procedures` and `finalize` though.

Monkatraz: also, it could allow this to be supported...
```ts
const MyServiceScaffold = ServiceSchema.scaffold({
  initializeState: () => ({ count: 1 }),
})

const myRPC = Procedure.rpc(MyServiceScaffold, {
  // ...
})
```
although the overloads would be bonkers. Maybe this would make more sense:
```ts
const myRPC = MyServiceScaffold.rpc({
  // ...
})
```
or maybe this is going too far, lol. I think people may occasionally want to construct procedures at the lowest indent level (if they're really complex), which is why I'm thinking about this.

router/services.ts
Monkatraz: ahh okay I get it, if you have a literal serialized state (in types) it lets you convert to (approximately) the service schema map that was serialized, albeit without state info

router/services.ts
jackyzha0: - `.scaffold` seems fine?
- `.procedures` for the actual procs makes sense to me

i think 3 levels of indent is ok, we can revisit if it turns out to be problematic?

Monkatraz: I will add that both me and Brady got annoyed at how much indentation procedure handlers were forced into, and we both focused on it quite a bit when trying to come up with APIs.

Monkatraz: Oh also we could just have both `ServiceSchema.define` and `ServiceSchema.scaffold`, small services get to have the nice ergonomics of `define` and huge services get to have the nice organization of `scaffold`.

jackyzha0: hm i would prefer not having two ways to construct services, i think scaffold isn't _that_ verbose

haha i was also thinking about 

```
.scaffold()
.rpc()
.procedure()
```

but then we are back in recursive type hell :P

bradymadden97: Plus one on indentation and composabity. I just find the builder pattern insanely hard to read lol but that might just be me. 

Also the (only) existing pid2 service has already grown to be like an 1000 line file :( so feels like a +1 towards an API that gives an easier way to draw boundaries and group things logically.

router/services.ts
```diff
@@ -251,18 +298,30 @@ export class ServiceSchema<
   >;
   // actual implementation
   static define(
```
jackyzha0, line 300: i wonder if, instead of overloading, we should just have `define` and `defineWithConfig` and have `define` just call `defineWithConfig` with the default config `{ initializeState: () => ({}) }`

router/result.ts
bradymadden97: Btw I discovered this change prevents us from returning `Type.Array` directly as output types, as it turns the array into a "readonly" array. So that's the tradeoff we're paying here by making literals a bit easier. I think we might want to consider reverting this, because having procedures return array payloads is pretty useful.

PR #84, opened 2024-04-19
reconnect on new proc after session end, session desync protocol error

jackyzha0:
## what changed

- rename `tryReconnecting` on `ClientTransport` to `reconnectOnConnectionDrop` as a more descriptive name
- add a test for 'client reconnects even after session grace'
- unnest `connectionRetryOptions` from `ClientTransport` options
- add `connectOnInvoke` option to handle cases where we want to reconnect after the session grace


tahazia-replit (approved):
LGTM!

PR #87, opened 2024-04-24
Force description types

bradymadden97:
Wrap and re-export typebox types to enforce us providing a "description" field for all parameters. We should now construct with `RiverType` instead of Typebox `Type`, to force us to provide a "description". This will flow through to the schema serialization.

Another idea would be to make `PayloadType` generic, which would potentially allow pid2 to enforce the description requirement instead of making it a river level construct. But frankly that seems just way more complicated, so I'm tempted to 🚢  this. Pulling it out to be a non-river level construct would mostly just to be more flexible to other non-replit users of river, which are not the priority at the moment. Relaxing the description requirement and making it generic in future releases would be backwards compat, so this is fine.

Fixed up all the tests and added one more for union becuase I wanted to make sure that worked as expected.

Not going to mark this release yet because I think there might be some other changes we want to make to serialization first.

### DIFF
--- a/__tests__/fixtures/services.ts
+++ b/__tests__/fixtures/services.ts
@@ -1,15 +1,21 @@
-import { Type } from '@sinclair/typebox';
+import { Type } from '../../types';
 import { ServiceSchema } from '../../router/services';
 import { Err, Ok } from '../../router/result';
 import { Observable } from './observable';
 import { Procedure } from '../../router';
 
-export const EchoRequest = Type.Object({
-  msg: Type.String(),
-  ignore: Type.Boolean(),
-  end: Type.Optional(Type.Boolean()),
-});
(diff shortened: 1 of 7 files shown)


__tests__/fixtures/services.ts
```diff
@@ -1,31 +1,45 @@
-import { Type } from '@sinclair/typebox';
+import { Type } from '../../types';
```
bradymadden97, line 1: i didn't know where to put this lol

types.ts
```diff
@@ -0,0 +1,158 @@
+import {
```
bradymadden97, line 1: probably need to export this somehow from the river pkg but was going to try and use this in pid2 and figure out exactly what needs to be exported

PR #88, opened 2024-04-25
Allow description types for procedures

bradymadden97:
This allows you to provide a description to procedures, which matches json schema. This will flow down to the serialized schema

### DIFF
--- a/router/procedures.ts
+++ b/router/procedures.ts
@@ -60,7 +60,7 @@ export interface RPCProcedure<
   input: I;
   output: O;
   errors: E;
-  description: string | undefined;
+  description?: string;
   handler(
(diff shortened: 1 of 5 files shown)


router/procedures.ts
```diff
@@ -60,6 +60,7 @@ export interface RPCProcedure<
   input: I;
   output: O;
   errors: E;
+  description: string | undefined;
```
jackyzha0, line 63: is there a reason we do this instead of just doing `description?: string`? is this to force people to consider a description

jackyzha0 (approved):

router/procedures.ts
bradymadden97: was fighting with the types but figured out a better way

PR #89, opened 2024-04-26
add `serialize()` to server

jackyzha0:
so we can avoid 

```
Object.entries(services).map(([name, value]) => ({
  name,
  ...value.serialize(),
})),
```

### DIFF
--- a/__tests__/serialize.test.ts
+++ b/__tests__/serialize.test.ts
@@ -4,6 +4,104 @@ import {
   FallibleServiceSchema,
   TestServiceSchema,
 } from './fixtures/services';
+import { createServer } from '../router/server';
+import { MockServerTransport } from './typescript-stress.test';
+
+describe('serialize server to jsonschema', () => {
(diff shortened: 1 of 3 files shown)


__tests__/serialize.test.ts
```diff
@@ -4,6 +4,104 @@ import {
...
+import { createServer } from '../router/server';
+import { MockServerTransport } from './typescript-stress.test';
+
+describe('serialize server to jsonschema', () => {
+  test('serialize basic server', () => {
+    const server = createServer(new MockServerTransport('mock'), {
+      test: TestServiceSchema,
+    });
+
+    expect(server.serialize()).toStrictEqual({
```
bradymadden97, line 16: FYI this is slightly different structure than we have right now I guess you're ok with that? I think python codegen might need to be fixed.

bradymadden97 (approved):
Makes sense. Kind of funny that the PR to avoid a snippet that's 4 lines is 120 lines hahaha

PR #91, opened 2024-04-30
Remove const modifier from result type

masad-frost:
Adding `const` generic modifier to our result functions makes it so that it's not possible to pass in arrays (possibly other things that are affected) to `Ok` (or `Err`).

For example

```ts
import {
  ServiceSchema,
  RiverUncaughtSchema,
  Procedure,
  Ok,
} from '@replit/river';
import { Type } from '@sinclair/typebox';

const SomethingWithArray = Type.Object({
  arr: Type.Array(Type.String()),
});

export const Service = ServiceSchema.define(
  {
    initializeState: () => ({}),
  },
  {
    initialize: Procedure.subscription({
      input: Type.Object({}),
      output: SomethingWithArray,
      errors: RiverUncaughtSchema,
      async handler(ctx, input, outputStream) {
        outputStream.push(
          Ok({
            arr: ['a', 'b'],
//         ^ ERROR, arr is not readonly
          }),
        );
      },
    }),
  },
);
```

I honestly don't understand what's going on here, and why we added the const modifier in the first place, so let me know if we should find another way to fix this.

Maybe we meant to do this instead? Produce a read-only result.

```
export function Ok<T, E>(payload: T): Result<const T, const E> {
  return {
    ok: true,
    payload,
  };
}
```

### DIFF
--- a/router/result.ts
+++ b/router/result.ts
@@ -46,14 +46,14 @@ export type Result<T, E> =
       payload: E;
     };
 
-export function Ok<const T, const E>(payload: T): Result<T, E> {
+export function Ok<T, E>(payload: T): Result<T, E> {
   return {
     ok: true,
     payload,
   };
 }
 
-export function Err<const T, const E>(error: E): Result<T, E> {
+export function Err<T, E>(error: E): Result<T, E> {
   return {
     ok: false,
     payload: error,


Monkatraz:
Here [is why we added it](https://github.com/replit/river/pull/76#discussion_r1559849550), quoting for posterity:
> If you have a strictly typed result like this:
> ```ts
> const MyOutput = Type.Object({ str: Type.Literal('foo') })
> type MyResult = Result<Static<typeof MyOutput>, never>
> ```
> then, counter-intuitively, this may not compile (without const generics):
> ```ts
> const myResult: MyResult = Ok({ str: 'foo' })
> ```
> this is because the object inside the `Ok` is typed broadly, like `{ str: string }`. It will compile if you do this though:
> ```ts
> const myResult: MyResult = Ok({ str: 'foo' } as const)
> ```
> which is annoying to have to type every time. The solution is to use const generics, which just does the `as const` for you (literally, that's all it does).

masad-frost:
https://github.com/microsoft/TypeScript/pull/55229

PR #92, opened 2024-04-30
Allow intersection types as payload, and nesting payloads

masad-frost:
Intersection of objects is useful when defining a schema.

There's also the issue that you can't nest unions (or intersections), ideally we write our types to allow infinite nesting as long as it bottoms out at objects, but that leads to infinite type checker recursion.
i.e. we can't do this
```ts
type PayloadType =
  | TObject
  | TUnion<Array<PayloadType>>
  | TIntersect<Array<PayloadType>>;
```

So wrote the nesting manual to allow 3 layers, hopefully that's enough for most schemas.

(courtesy of @Monkatraz)

### DIFF
--- a/router/procedures.ts
+++ b/router/procedures.ts
@@ -28,10 +28,37 @@ export type ValidProcType =
   // Bidirectional stream (potentially preceded by an initialization message) (n:n).
   | 'stream';
 
+type PayloadType0 =
+  | TObject
+  | TUnion<Array<TObject>>
+  | TIntersect<Array<TObject>>;
+
+type PayloadType1 =
+  | TObject
+  | TUnion<Array<PayloadType0>>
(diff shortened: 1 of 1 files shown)


router/procedures.ts
```diff
@@ -28,10 +35,37 @@ export type ValidProcType =
...
+
+type PayloadType1 =
+  | TObject
+  | TUnion<Array<PayloadType0>>
+  | TIntersect<Array<PayloadType0>>;
+
+type PayloadType2 =
+  | TObject
+  | TUnion<Array<PayloadType1>>
+  | TIntersect<Array<PayloadType1>>;
```
jackyzha0, line 51: can we use https://github.com/sinclairzx81/typebox#recursive-types for this?

masad-frost:
#93

PR #93, opened 2024-05-01
loosen payload type requirement

jackyzha0:
supersedes https://github.com/replit/river/pull/92

### DIFF
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -19,6 +19,7 @@ import {
   TestServiceSchema,
   UploadableServiceSchema,
   OrderingServiceSchema,
+  NonObjectSchemas,
 } from './fixtures/services';
 import { UNCAUGHT_ERROR } from '../router/result';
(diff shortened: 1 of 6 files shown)


masad-frost (approved):
【﻿ｂａｓｅｄ】


image

masad-frost:
Probably nothing wrong with non-object schemas

jackyzha0:
> Probably nothing wrong with non-object schemas

yeah we'll probably want to make sure codegen still works with this but otherwise seems fine 🤷

masad-frost:
Only thing is that we run into the off chance of something inventing a type that's not serializable to json schema

jackyzha0:
> Only thing is that we run into the off chance of something inventing a type that's not serializable to json schema

i believe everything `TSchema` represents can be serialized into json schema? 👀

bradymadden97:
This will be better for agent RPCs too. Less tokens, less complexity building the types, etc.

PR #98, opened 2024-05-03
RFC: handle bad messages

masad-frost:
I ran into a situation where my pid2 didn't have a procedure I was using on the client, and things just hanged :(.

Just an RFC, no implementation.

Left comments and suggestions. But here's a summary:
- We add a new error `INVALID_REQUEST`. We use it to respond to bad requests
- We start aggressively disconnecting clients that are misbehave

### DIFF
--- a/router/result.ts
+++ b/router/result.ts
@@ -28,10 +28,12 @@ export type RiverError =
 
 export const UNCAUGHT_ERROR = 'UNCAUGHT_ERROR';
 export const UNEXPECTED_DISCONNECT = 'UNEXPECTED_DISCONNECT';
+export const INVALID_REQUEST = 'INVALID_REQUEST';
 export const RiverUncaughtSchema = Type.Object({
   code: Type.Union([
     Type.Literal(UNCAUGHT_ERROR),
     Type.Literal(UNEXPECTED_DISCONNECT),
+    Type.Literal(INVALID_REQUEST),
   ]),
   message: Type.String(),
 });
--- a/router/server.ts
+++ b/router/server.ts
@@ -119,8 +119,13 @@ class RiverServer<Services extends ServiceSchemaMap> {
       log?.info(
         `${this.transport.clientId} -- got msg with destination that isn't the server, ignoring`,
       );
+
+      // This is a bad client
+      // TODO disconnect session
       return;
     }
+
+    // TODO check if TransportMessageSchema is satisfied
 
     let procStream = this.streamMap.get(message.streamId);
     const isInitMessage = !procStream;
@@ -167,14 +172,20 @@ class RiverServer<Services extends ServiceSchemaMap> {
         `${this.transport.clientId} -- can't create a new procedure stream from a message that doesn't have the stream open bit set`,
       );
       log?.debug(` -> ${JSON.stringify(message)}`);
+
+      // This is a bad client
+      // TODO: disconnect session.
       return;
     }
 
     if (!message.procedureName || !message.serviceName) {
-      log?.warn(
+      log?.error(
         `${this.transport.clientId} -- missing procedure or service name in stream open message`,
       );
       log?.debug(` -> ${JSON.stringify(message)}`);
+
+      // This is a bad client
+      // TODO: disconnect session.
       return;
     }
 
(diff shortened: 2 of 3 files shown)


router/result.ts
```diff
@@ -28,10 +28,12 @@ export type RiverError =
 
 export const UNCAUGHT_ERROR = 'UNCAUGHT_ERROR';
 export const UNEXPECTED_DISCONNECT = 'UNEXPECTED_DISCONNECT';
+export const INVALID_REQUEST = 'INVALID_REQUEST';
 export const RiverUncaughtSchema = Type.Object({
   code: Type.Union([
     Type.Literal(UNCAUGHT_ERROR),
     Type.Literal(UNEXPECTED_DISCONNECT),
+    Type.Literal(INVALID_REQUEST),
```
masad-frost, line 36: This is probably protocol major bump

router/server.ts
```diff
@@ -119,9 +119,14 @@ class RiverServer<Services extends ServiceSchemaMap> {
       log?.info(
         `${this.transport.clientId} -- got msg with destination that isn't the server, ignoring`,
       );
+
+      // This is a bad client
+      // TODO disconnect session
       return;
     }
 
+    // TODO check if TransportMessageSchema is satisfied
```
masad-frost, line 128: This is already checked in the transport, nvm

bradymadden97:
What's the benefit to disconnecting clients that are sending invalid requests? Over just responding to them w/ "Invalid Request" Are there analogs to that in other protocols? Thinking of like LSP I don't believe it ends connections after bad messages, it just responds w/ errors to those messages.

router/server.ts
```diff
@@ -167,21 +172,32 @@ class RiverServer<Services extends ServiceSchemaMap> {
...
     }
 
     if (!(message.serviceName in this.services)) {
       log?.warn(
         `${this.transport.clientId} -- couldn't find service ${message.serviceName}`,
       );
+
+      // Either a bad client or backwards incompatible mismatched versions between client and server
+      // TODO: disconnect session
+      // TODO alternative 1: respond with an error message
```
masad-frost, line 199: luis says this one

router/server.ts
```diff
@@ -191,6 +207,11 @@ class RiverServer<Services extends ServiceSchemaMap> {
       log?.warn(
         `${this.transport.clientId} -- couldn't find a matching procedure for ${message.serviceName}.${message.procedureName}`,
       );
+
+      // Either a bad client or backwards incompatible mismatched versions between client and server
+      // TODO: disconnect session
+      // TODO alternative 1: respond with an error message
```
masad-frost, line 213: luis says this one

router/server.ts
```diff
@@ -167,21 +172,32 @@ class RiverServer<Services extends ServiceSchemaMap> {
...
 
     if (!message.procedureName || !message.serviceName) {
-      log?.warn(
+      log?.error(
         `${this.transport.clientId} -- missing procedure or service name in stream open message`,
       );
       log?.debug(` -> ${JSON.stringify(message)}`);
+
+      // This is a bad client
+      // TODO: disconnect session.
```
masad-frost, line 188: send back an error instead

router/server.ts
```diff
@@ -167,21 +172,32 @@ class RiverServer<Services extends ServiceSchemaMap> {
         `${this.transport.clientId} -- can't create a new procedure stream from a message that doesn't have the stream open bit set`,
       );
       log?.debug(` -> ${JSON.stringify(message)}`);
+
+      // This is a bad client
+      // TODO: disconnect session.
```
masad-frost, line 177: this could be a bad server too

masad-frost:
> What's the benefit to disconnecting clients that are sending invalid requests? Over just responding to them w/ "Invalid Request" Are there analogs to that in other protocols? Thinking of like LSP I don't believe it ends connections after bad messages, it just responds w/ errors to those messages.

Depends on how gracefully someone can recover from an error, certain errors can mean every future message by the client will be wrong (e.g. a wrong sequence number), we don't want to keep the client lying around spamming the server with errors.

bradymadden97:
> Depends on how gracefully someone can recover from an error, certain errors can mean every future message by the client will be wrong (e.g. a wrong sequence number), we don't want to keep the client lying around spamming the server with errors.

Sure, I think we could separate those out though, no? Like a malformed payload doesn't mean every subsequent payload will be malformed.

masad-frost:
closing for now, we're kind of blocked on making the close functionality more robust as described in #111

PR #99, opened 2024-05-03
add disposable to stream

jackyzha0:


Monkatraz (approved):
lgtm, honestly I forgot this wasn't a feature since it seems so natural

PR #103, opened 2024-05-07
structured logging

jackyzha0:
## Why

- we'd like to be able to be more intelligent about what we log and how
- this will enable things like selectively forwarding server logs to the browser, adding metadata fields to datadog, etc.

## What changed

- changed the logger interface
- rely on metadata obj rather than string formatting

## Versioning


logging/index.ts
```diff
@@ -1,69 +1,89 @@
...
+interface LogMessage {
+  level: keyof typeof LoggingLevels;
+  message: string;
+  metadata?: MessageMetadata;
+}
+
 type LoggingLevel = keyof typeof LoggingLevels;
-export type Logger = {
+
+export type LogFn = (msg: LogMessage) => void;
```
masad-frost, line 29: Most logging libraries do `(msg: string, data: Record<string, unknown>)`, we should probably stick to that, most have specific functions for the level, but I've seen logdna and datadog also provide a more generic function that accepts a third argument for the level.

logging/index.ts
```diff
@@ -1,69 +1,89 @@
...
- * Sets the minimum logging level for the logger.
- * @param level - The minimum logging level to set.
+ * The global River logger instance.
  */
-export function setLevel(level: LoggingLevel) {
-  if (log) {
-    log.minLevel = level;
-  }
+export let log: Logger | undefined = undefined;
+export function bindLogger(fn: LogFn, level?: LoggingLevel) {
```
masad-frost, line 86: I feel like we should accept a logger shape here `LogFn | { error: LogFn, warn: LogFn, info: LogFn, debug: LogFn }`, I think it'd be much easier for us with datadog.

logging/index.ts
```diff
@@ -1,69 +1,89 @@
...
+  warn: '\u001b[33m',
+  error: '\u001b[31m',
+};
+
+export const coloredStringLogger: LogFn = (msg) => {
+  const color = colorMap[msg.level];
+  return `${color}${msg.level}\u001b[0m ${msg.message}`;
+};
+
+export const jsonLogger: LogFn = JSON.stringify;
```
masad-frost, line 80: how are these loggers? They just format messages.

logging/index.ts
```diff
@@ -1,69 +1,89 @@
...
 /**
- * Sets the minimum logging level for the logger.
- * @param level - The minimum logging level to set.
+ * The global River logger instance.
  */
-export function setLevel(level: LoggingLevel) {
-  if (log) {
-    log.minLevel = level;
-  }
+export let log: Logger | undefined = undefined;
```
masad-frost, line 85: It seems like the logger isn't meant for re-consumptions, just internal use, we should try to remove it from public interface

masad-frost (changes requested):

masad-frost (approved):

logging/index.ts
```diff
@@ -1,69 +1,91 @@
...
+  }
 
-/**
- * Unbinds the logger so subsequent logs do not call
- * the write callback previously provided to {@link bindLogger}
- */
-export function unbindLogger() {
-  log = undefined;
+  info(msg: string, metadata?: MessageMetadata) {
+    if (LoggingLevels[this.minLevel] > LoggingLevels.info) return;
```
bradymadden97, line 47: lol why not just write a normal conditional wrapping this.output? we don't have code like this really anywhere else at replit

transport/message.ts
```diff
@@ -103,7 +103,7 @@ export const OpaqueTransportMessageSchema = TransportMessageSchema(
  *   contain any payload that is relevant to the application so should not be delivered.
  * @template Payload The type of the payload.
  */
-export interface TransportMessage<Payload = Record<string, unknown>> {
+export interface TransportMessage<Payload = unknown> {
```
bradymadden97, line 106: why were all of these changed, i don't think i understand

logging/index.ts
```diff
@@ -1,69 +1,91 @@
...
 const LoggingLevels = {
   debug: -1,
   info: 0,
   warn: 1,
   error: 2,
 } as const;
-
 type LoggingLevel = keyof typeof LoggingLevels;
+
+export type LogFn = (
```
bradymadden97, line 12: what would be helpful, to unblock https://github.com/replit/repl-it-web/pull/44111, would be to provide some information that a particular log is coming from within river, so we can not forward it. otherwise this doesn't really help us filter out and avoid the infinite loops

transport/message.ts
jackyzha0: this is just misc type cleanup following https://github.com/replit/river/commit/83af11d57ea8c5a8aaca52d254425117159180e4

logging/index.ts
jackyzha0: just choosing to early return here

bradymadden97 (approved):
nice!!

PR #105, opened 2024-05-08
export some more things, accept broader log interface

jackyzha0:
## Why

- old required you to pass in an explicit BaseLogger, let's just accept anything that satisfies Logger

## What changed

- accept Logger instead of BaseLogger
- export Logger interface

## Versioning

### DIFF
--- a/logging/log.ts
+++ b/logging/log.ts
@@ -29,7 +29,7 @@ export type MessageMetadata = Record<string, unknown> &
     partialTransportMessage: Partial<PartialTransportMessage>;
   }>;
 
-class BaseLogger {
+class BaseLogger implements Logger {
   minLevel: LoggingLevel;
   private output: LogFn;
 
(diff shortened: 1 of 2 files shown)


logging/log.ts
```diff
@@ -75,29 +75,29 @@ const colorMap = {
...
+): Logger | undefined {
   if (!fn) {
     log = undefined;
     return;
   }
 
-  if (fn instanceof BaseLogger) {
-    log = fn;
-    return fn;
+  if (fn instanceof Function) {
```
masad-frost, line 96:
```suggestion
  if (typeof fn === 'function') {
```

It's generally considered a safer choice and wont have issues with cross context functions.

logging/log.ts
```diff
@@ -75,29 +75,29 @@ const colorMap = {
 };
 
 export const coloredStringLogger: LogFn = (msg, _ctx, level) => {
```
masad-frost, line 77: Maybe use default arg?

PR #109, opened 2024-05-09
Don't expose `Pushable` from client

masad-frost:
## Why

While working on #98, I noticed that we can make our API more consistent when it comes to closing if we always expose a close method, especially for streams you can both call `input.end` and `close` to end the stream, 2 ways of doing things is not good, and the behavior of `input.end` is not self-explanatory as it is with `close`. 

Note: This uncovered a larger discussion/issue about our protocol, I found in the implementation that sometimes the server can initiate a close and it expects the client to send a close back. This is not documented in our protocol. We're considering doing half closes, TBD https://replit.slack.com/archives/C05GZ4NRSJ3/p1715235702921009. We'll address that in a different PR, but this PR essentially makes it so that server close requests are half-closes, some tests failing in this PR rely on client responding, so please ignore for now.

## What changed

We never expose `Pushable` on the client side, what were previously input `Pushable` is now a simple "send" callback. Our "receives" were already exposed as async iterables, not `Pushable`, so we're covered there.

- `upload` returns `[send, finalize]`
  - `send` is a simple callback
  - `finalize` finalizes the request and returns a promise
  - it is now an error to send a message after calling finalize (previously you could call `input.end` followed by `input.push`)
  - receiving a reponse before we send a finalize request is now considered an error, more closely matching our protocol docs
```
client: > --  - {
server:          <
```
- `stream` now returns `send` function as a first argument instead of a pushable
  -  `close` function is the only way to close the stream
  - it is now an error to send a message after calling close (previously you could call `input.end` followed by `input.push`)

there are some other minor changes like re-using the `send` function for `init` messages.

I have some test changes, but decided not to check them since they don't make sense until we figure out what to do about server initiated closes.

## Versioning

### DIFF
--- a/router/client.ts
+++ b/router/client.ts
@@ -11,7 +11,6 @@ import {
   InstantiatedServiceSchemaMap,
 } from './services';
 import { pushable } from 'it-pushable';
-import type { Pushable } from 'it-pushable';
 import {
   OpaqueTransportMessage,
   ControlFlags,
@@ -32,6 +31,13 @@ export type AsyncIter<T> = AsyncGenerator<T, T>;
 /**
  * A helper type to transform an actual service type into a type
  * we can case to in the proxy.
+ *
+ * If you end up here in "click to definition", you probably want
+ * to click to definition on the procedure not the procedure type,
+ * which will take you to the schema definition that contains the
+ * input, output, and error types.
+ * e.g. client.someService.someProcedure.rpc
+ *                         ^^^^^^^^^^^^^ Click here
  * @template Router - The type of the Router.
  */
 type ServiceClient<Router extends AnyService> = {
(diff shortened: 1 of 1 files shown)


router/client.ts
```diff
@@ -304,49 +306,40 @@ function handleStream(
...
+    if (firstMessage) {
+      m.serviceName = serviceName;
+      m.procedureName = procedureName;
+      m.controlFlags |= ControlFlags.StreamOpenBit;
+      firstMessage = false;
     }
 
-    // after ending input stream, send a close message to the server
-    if (!healthyClose) return;
-    transport.sendCloseStream(serverId, streamId);
```
masad-frost, line 346: this used to send a close response back to the server, relating to the big note in the description.

masad-frost:
Closing this in favor of efforts starting in  #118

PR #110, opened 2024-05-09
[WS-3468] River-side changes for Option A of identity work

Monkatraz:
## Why
We want identity/auth support in River.

## What changed
Made it so that client and server transports can extend the data in the handshake. An example works best:
```ts
// add our handshake types
declare module "@replit/river" {
  interface HandshakeRequestMetadata {
    username: string
    secret: string
  }
  
  interface ParsedHandshakeMetadata {
    username: string
  }
}

const clientTransport = new MyClientTransport('client', {
  handshake: {
    // runtime validates the result of `get`
    schema: Type.Object({
      username: Type.String(),
      secret: Type.string(),
    }),
    get: async () => {
      return await getClientHandshakeMetadata()
    }
  }
})

const serverTransport = new MyServerTransport('server', {
  handshake: {
    // runtime validates the metadata in the handshake request
    requestSchema: Type.Object({
      username: Type.String(),
      secret: Type.string(),
    }),

    // runtime validates the parsed metadata
    parsedSchema: Type.Object({
      username: Type.String(),
    }),
    
    // session will not be null if this is a reconnect
    parse: async (metadata, session, isReconnect) => {
      // reject connection
      if (await isInvalidAuth(metadata)) {
        return false
      }
      
      // reject reconnection
      if (isReconnect && session.metadata?.username !== metadata.username) {
        return false
      }
      
      // stored on sessions
      return { username: metadata.username }
    }
  }
})

const MyService = ServiceSchema.define({
  foo: Procedure.rpc({
    // ...
    async handler(ctx, _input) {
      // guaranteed to not be undefined
      const username = ctx.session.metadata.username
      return Ok(`your name is ${username}`)
    }
  })
})
```

I think there are some open questions about the names of things and the ergonomics of this, but this should be everything we need.

## Versioning

### DIFF
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -34,7 +34,7 @@ describe.each(testMatrix())(
   async ({ transport, codec }) => {
     const opts = { codec: codec.codec };
     const { getClientTransport, getServerTransport, cleanup } =
-      await transport.setup(opts);
+      await transport.setup({ client: opts, server: opts });
     afterAll(cleanup);
 
     test('rpc', async () => {
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -22,13 +22,15 @@ import {
   NonObjectSchemas,
   SchemaWithDisposableState,
 } from './fixtures/services';
-import { UNCAUGHT_ERROR } from '../router/result';
+import { Ok, UNCAUGHT_ERROR } from '../router/result';
 import {
   advanceFakeTimersBySessionGrace,
   testFinishesCleanly,
   waitFor,
 } from './fixtures/cleanup';
 import { testMatrix } from './fixtures/matrix';
+import { Type } from '@sinclair/typebox';
+import { Procedure, ServiceSchema } from '../router';
 
 describe.each(testMatrix())(
   'client <-> server integration test ($transport.name transport, $codec.name codec)',
@@ -652,6 +654,89 @@ describe.each(testMatrix())(
       // test
       await server.close();
       expect(dispose).toBeCalledTimes(1);
+    });
+  },
+);
+
+describe.each(testMatrix())(
+  'client <-> server with handshake tests ($transport.name transport, $codec.name codec)',
+  async ({ transport, codec }) => {
+    const requestSchema = Type.Object({
+      data: Type.String(),
+    });
+
+    const parsedSchema = Type.Object({
+      data: Type.String(),
+      extra: Type.Number(),
+    });
+
+    const { getClientTransport, getServerTransport, cleanup } =
+      await transport.setup({
+        client: {
+          codec: codec.codec,
+          handshake: {
+            schema: requestSchema,
+            get: () => ({ data: 'foobar' }),
+          },
+        },
+
+        server: {
+          codec: codec.codec,
+          handshake: {
+            requestSchema,
+            parsedSchema,
+            parse: (metadata) => {
+              return {
+                // @ts-expect-error we haven't extended the interface
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                data: metadata.data,
+                extra: 42,
+              };
+            },
+          },
+        },
+      });
+
+    afterAll(cleanup);
+
+    test('procedure can use metadata', async () => {
+      // setup
+      const clientTransport = getClientTransport('client');
+      const serverTransport = getServerTransport();
+      const services = {
+        test: ServiceSchema.define({
+          getData: Procedure.rpc({
+            input: Type.Object({}),
+            output: Type.Object({
+              data: Type.String(),
+              extra: Type.Number(),
+            }),
+            handler: async (ctx) => {
+              // we haven't extended the interface, so we need to suppress the error
+              // with a cast
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              return Ok({ ...ctx.session.metadata } as any);
+            },
+          }),
+        }),
+      };
+      const server = createServer(serverTransport, services);
+      const client = createClient<typeof services>(
+        clientTransport,
+        serverTransport.clientId,
+      );
+      onTestFinished(async () => {
+        await testFinishesCleanly({
+          clientTransports: [clientTransport],
+          serverTransport,
+          server,
+        });
+      });
+
+      // test
+      const result = await client.test.getData.rpc({});
+      assert(result.ok);
+      expect(result.payload).toStrictEqual({ data: 'foobar', extra: 42 });
     });
   },
 );
--- a/router/context.ts
+++ b/router/context.ts
@@ -1,3 +1,10 @@
+import {
+  Connection,
+  ParsedHandshakeMetadata,
+  Session,
+  TransportClientId,
+} from '../transport';
+
 /**
  * The context for services/procedures. This is used only on
  * the server.
@@ -18,9 +25,6 @@
  * }
  * ```
  */
-
-import { Connection, Session, TransportClientId } from '../transport';
-
 /* eslint-disable-next-line @typescript-eslint/no-empty-interface */
 export interface ServiceContext {}
 
@@ -34,5 +38,5 @@ export type ServiceContextWithTransportInfo<State> = ServiceContext & {
   to: TransportClientId;
   from: TransportClientId;
   streamId: string;
-  session: Session<Connection>;
+  session: Session<Connection> & { handshakeMetadata: ParsedHandshakeMetadata };
 };
--- a/router/context.ts
+++ b/router/context.ts
@@ -38,5 +38,5 @@ export type ServiceContextWithTransportInfo<State> = ServiceContext & {
   to: TransportClientId;
   from: TransportClientId;
   streamId: string;
-  session: Session<Connection> & { handshakeMetadata: ParsedHandshakeMetadata };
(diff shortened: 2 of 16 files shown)


transport/message.ts
```diff
@@ -14,6 +15,102 @@ export const enum ControlFlags {
...
+   */
+  requestSchema: TSchema;
+
+  /**
+   * Schema for the transformed metadata that is then associated with the
+   * client's session.
+   *
+   * Needs to match {@link ParsedHandshakeMetadata}.
+   */
+  parsedSchema: TSchema;
```
Monkatraz, line 93: Do we need this?

transport/session.ts
```diff
@@ -122,6 +123,13 @@ export class Session<ConnType extends Connection> {
    */
   advertisedSessionId?: string;
 
+  /**
+   * The metadata for this session, as parsed from the handshake.
+   *
+   * Will only ever be populated on the server side.
+   */
+  handshakeMetadata?: ParsedHandshakeMetadata;
```
Monkatraz, line 131: Open to better names here, it's kind of unergonomic

transport/transport.test.ts
```diff
@@ -728,3 +738,525 @@ describe.each(testMatrix())(
     });
   },
 );
+
+describe.each(testMatrix())(
```
Monkatraz, line 742: Each of these tests feel unnecessarily gigantic, not sure what to do about it (maybe this is just fine?)

transport/message.ts
```diff
@@ -14,6 +15,102 @@ export const enum ControlFlags {
...
+ * Options for extending the client handshake process.
+ */
+export interface ClientHandshakeOptions {
+  /**
+   * Schema for the metadata that the client sends to the server
+   * during the handshake.
+   *
+   * Needs to match {@link HandshakeRequestMetadata}.
+   */
+  schema: TSchema;
```
Monkatraz, line 67: Do we care about the client validating the metadata type?

router/context.ts
```diff
@@ -34,5 +38,5 @@ export type ServiceContextWithTransportInfo<State> = ServiceContext & {
   to: TransportClientId;
   from: TransportClientId;
   streamId: string;
-  session: Session<Connection>;
+  session: Session<Connection> & { handshakeMetadata: ParsedHandshakeMetadata };
```
Monkatraz, line 41: This + a check in the server code means we guarantee that we always have metadata in procedures

masad-frost: wonder if we put this directly on the service context

transport/transport.ts
```diff
@@ -835,7 +921,12 @@ export abstract class ServerTransport<
         reason,
       });
       conn.send(this.codec.toBuffer(responseMsg));
-      log?.warn(`${reason}: ${JSON.stringify(parsed)}`, {
+      // note: do _not_ log the payload, it may contain secrets
```
masad-frost, line 924: nice

transport/transport.ts
```diff
@@ -835,7 +921,12 @@ export abstract class ServerTransport<
...
+      const logData =
+        typeof parsed.payload === 'object' && parsed.payload !== null
```
masad-frost, line 926:
```suggestion
        typeof parsed.payload === 'object'
```

we already check falsey values above :shrug:

transport/transport.ts
```diff
@@ -863,7 +954,79 @@ export abstract class ServerTransport<
...
+          this.clientId,
+          parsed.from,
+          { ok: false, reason },
+        );
+        conn.send(this.codec.toBuffer(responseMsg));
+        // note: do _not_ log the metadata, it may contain secrets
+        log?.warn(`received malformed handshake metadata from ${parsed.from}`, {
+          clientId: this.clientId,
+          connId: conn.debugId,
+        });
```
masad-frost, line 979: Wonder if we log `Value.Errors` here with redacted `value` fields.

transport/transport.ts
```diff
@@ -863,7 +954,79 @@ export abstract class ServerTransport<
...
+        this.protocolError(ProtocolError.HandshakeFailed, reason);
+        return false;
+      }
+
+      const parsedMetadata = await this.options.handshake.parse(
+        parsed.payload.metadata as HandshakeRequestMetadata,
+        session ?? null,
```
masad-frost, line 986: wonder if we always pass `session` (i.e. change above to `getOrCreateSession` and make `getOrCreateSession` tell us if this is existing or new) and add a boolean parameter to `handshake.parse` for whether it's a new session or an existing one.

transport/transport.ts
```diff
@@ -863,7 +954,79 @@ export abstract class ServerTransport<
...
+      // check that the parsed metadata is the correct shape
+      if (!Value.Check(this.options.handshake.parsedSchema, parsedMetadata)) {
+        const reason = 'failed to parse handshake metadata';
+        const responseMsg = handshakeResponseMessage(
+          this.clientId,
+          parsed.from,
+          { ok: false, reason },
+        );
+        conn.send(this.codec.toBuffer(responseMsg));
+        // note: do _not_ log the metadata, it may contain secrets
```
masad-frost, line 1015: Same as above re `Value.Errors`

transport/message.ts
masad-frost: doesn't hurt probably

masad-frost (approved):
Nice ty ty.

Would be interesting to see how we can/should expose those to codegen for other clients.

transport/transport.test.ts
masad-frost:
```suggestion
describe.each([testMatrix()[0]])(
```
?
lol

masad-frost: runs fast enough, we can optimize tests later

__tests__/e2e.test.ts
masad-frost, line 1: Probably a good idea to add an `e2e` test where we have a service with procedures that read the metadata and sends it back to the client or something.

transport/transport.test.ts
Monkatraz: Meant more that the tests themselves are literally huge in line count

masad-frost: I think it's fine.

https://testing.googleblog.com/2019/12/testing-on-toilet-tests-too-dry-make.html?m=1

transport/transport.ts
Monkatraz: As far as I can tell, we don't log anything like `Value.Errors`.

jackyzha0: we could 👀

transport/transport.ts
masad-frost: Sweet, thanks. Yeah looks good

router/server.ts
```diff
@@ -262,6 +262,15 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
       );
     };
 
+    // by this point, our sessions should always have their handshake metadata
+    if (session.metadata === undefined) {
+      log?.error(
+        `session ${message.from} doesn't have handshake metadata, can't proceed with procedure`,
```
jackyzha0, line 268:
```suggestion
        `(invariant violation) session doesn't have handshake metadata`,
```

we already log the from in the `session.loggingMetadata`

transport/message.ts
jackyzha0: do you think maybe we could do https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces-with-classes-functions-and-enums and the handshake process a class? and maybe put the schema as a static member of the class so that way we can infer HandshakeRequestMetadata via `Static<T>`

transport/transport.test.ts
```diff
@@ -728,3 +738,527 @@ describe.each(testMatrix())(
...
+      onTestFinished(cleanup);
+
+      const clientTransport = getClientTransport('client');
+      const serverTransport = getServerTransport();
+
+      await waitFor(() => {
+        expect(serverTransport.sessions.size).toBe(1);
+        expect(get).toHaveBeenCalledTimes(1);
+        expect(parse).toHaveBeenCalledTimes(1);
+      });
```
jackyzha0, line 792: damn lol didnt realize you could wrap waitFor like this 🤦

transport/transport.test.ts
```diff
@@ -728,3 +738,527 @@ describe.each(testMatrix())(
...
+      const { getClientTransport, getServerTransport, cleanup } =
+        await transport.setup(opts);
+      onTestFinished(cleanup);
+
+      const clientTransport = getClientTransport('client');
+      const serverTransport = getServerTransport();
+
+      const clientHandshakeFailed = vi.fn();
+      const clientHandshakeFailedHandler = (evt: EventMap['protocolError']) => {
+        if (evt.type === 'handshake_failed') {
```
jackyzha0, line 938: should use the const enum here `ProtocolError.HandshakeFailed`

transport/transport.test.ts
Monkatraz: It's kind of bad for reporting errors, sometimes it will report what failed here sometimes it just says it timed out 🤷

transport/message.ts
Monkatraz: My gut is that it's maybe possible but probably pretty goofy to use, since it's already kind of an esoteric feature to do a namespace/interface merge with something like a class. I'll look more into it, but just to get the ball rolling I'll move forward with this system

PR #112, opened 2024-05-09
Support disposing from service state

bradymadden97:
## Why

Services might set up things that need to be disposed. We can support this in a non breaking way by not _requiring_ service context to implement `Disposable`, but if it does, we can call `[Symbol.dispose]` when the server is closing.

This follows the explicit resource management proposal https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html 

## Testing

CI

## Versioning

### DIFF
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -641,10 +641,6 @@ describe.each(testMatrix())(
         disposable: SchemaWithDisposableState(dispose),
       };
       const server = createServer(serverTransport, services);
-      const client = createClient<typeof services>(
-        clientTransport,
-        serverTransport.clientId,
-      );
       onTestFinished(async () => {
         await testFinishesCleanly({
           clientTransports: [clientTransport],
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -20,6 +20,7 @@ import {
   UploadableServiceSchema,
   OrderingServiceSchema,
   NonObjectSchemas,
+  SchemaWithDisposableState,
 } from './fixtures/services';
 import { UNCAUGHT_ERROR } from '../router/result';
 import {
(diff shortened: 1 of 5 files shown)


router/procedures.ts
```diff
@@ -62,7 +62,9 @@ export interface RPCProcedure<
   errors: E;
   description?: string;
   handler(
-    context: ServiceContextWithTransportInfo<State>,
+    context: ServiceContextWithTransportInfo<
```
bradymadden97, line 65: This prevents exposing the dispose method to handlers, which should not really be accessing it.

__tests__/e2e.test.ts
```diff
@@ -630,5 +631,33 @@ describe.each(testMatrix())(
...
+        clientTransport,
+        serverTransport.clientId,
+      );
+      onTestFinished(async () => {
+        await testFinishesCleanly({
+          clientTransports: [clientTransport],
+          serverTransport,
+          server,
+        });
+        expect(dispose).toBeCalledTimes(1);
```
jackyzha0, line 654: probably antipattern to expect in onTestFinished,

we should manually close the server and expect it to be closed

router/procedures.ts
jackyzha0: we can probably do this omit in `ServiceContextWithTransportInfo` as this is just a handler-level construction

router/server.ts
```diff
@@ -161,6 +161,15 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
     this.transport.removeEventListener('message', this.onMessage);
     this.transport.removeEventListener('sessionStatus', this.onSessionStatus);
     await Promise.all([...this.streamMap.keys()].map(this.cleanupStream));
+
+    for (const [_, context] of this.contextMap.entries()) {
```
jackyzha0, line 165:
```suggestion
    for (const context of this.contextMap.values()) {
```

bradymadden97:
the issue is i don't think we actually call `server.close()` anywhere. i guess eventually if we move to `using server` (following the explicit resource management proposal) and update our `RiverServer` to implement `Disposable`, then all the cleanups will flow down. 

for now i suppose everything just gets killed unsafely when the server dies.

jackyzha0 (approved):

__tests__/e2e.test.ts
bradymadden97: haha but that's what onTestFinished is doing internally? it has a bunch of expects. just hidden behind 3 layers of abstraction :)

bradymadden97: erm, testFinishesCleanly

jackyzha0: yeah but those are invariant checks, it's not the core logic of the test if that makes sense

PR #117, opened 2024-05-10
Add generic `Output<>` type for retrieving output message type on client

bradymadden97:
## Why

You may want to use the type of the output message on the client, like in helper functions, without pulling the typebox schema into a location that you can export and share across client and server. Instead, river can export a generic type helper you can use instead that derives the type from the client type.

As the docblock says, you can retrieve the `Result` type for the output message of any procedure by doing the following:
```typescript
type Message = Output<typeof client, 'serviceName', 'procedureName'>
```

## Test plan

Works with all four procedure types!
Screen Shot 2024-05-10 at 11 49 38 AM
Screen Shot 2024-05-10 at 11 45 42 AM
Screen Shot 2024-05-10 at 11 45 34 AM
Screen Shot 2024-05-10 at 11 45 26 AM

## Versioning

### DIFF
--- a/router/result.ts
+++ b/router/result.ts
@@ -7,6 +7,7 @@ import {
   TUnion,
   Type,
 } from '@sinclair/typebox';
+import { Client } from './client';
 
 type TLiteralString = TLiteral<string>;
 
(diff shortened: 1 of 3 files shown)


router/result.ts
```diff
@@ -59,3 +60,69 @@ export function Err<const T, const E>(error: E): Result<T, E> {
     payload: error,
   };
 }
+
+/**
+ * Refine a {@link Result} type to its returned value.
+ */
+export type ResultOk<R> = R extends Result<infer __T, infer __E> & {
```
jackyzha0, line 67: nit: i think `ResultUnwrapOk` and `ResultUnwrapErr` make slightly more sense here

router/result.ts
```diff
@@ -59,3 +60,69 @@ export function Err<const T, const E>(error: E): Result<T, E> {
...
+  ? A
+  : never;
+
+/**
+ * Refine a {@link Result} type to its error value.
+ */
+export type ResultErr<R> = R extends Result<infer __T, infer __E> & {
+  ok: false;
+  payload: infer A;
+}
```
jackyzha0, line 80: why doesnt

```suggestion
export type ResultErr<R> = R extends Result<infer __T, infer E> ? E : never;
export type ResultOk<R> = R extends Result<infer T, infer __E> ? T : never;
```

work? just curious

router/result.ts
```diff
@@ -59,3 +60,69 @@ export function Err<const T, const E>(error: E): Result<T, E> {
...
+        }
+      ? Awaited<ReturnType<SubscriptionHandler>> extends [
+          AsyncGenerator<infer SubscriptionOutputMessage>,
+          infer __SubscriptionCloseHandle,
+        ]
+        ? SubscriptionOutputMessage
+        : never
+      : never
+    : never
+  : never;
```
jackyzha0, line 128: they've played us for absolute fools

image

router/result.ts
```diff
@@ -59,3 +60,69 @@ export function Err<const T, const E>(error: E): Result<T, E> {
...
+
+/**
+ * Retrieve the output type for a procedure, represented as a {@link Result}
+ * type.
+ * Example:
+ * ```
+ * type Message = Output<typeof client, 'serviceName', 'procedureName'>
+ * ```
+ */
+export type Output<
```
masad-frost, line 92: Can we use these in one of the tests or just a random test file so that if they break we get a signal before a broken version out?

masad-frost (approved):

masad-frost:
ping dhh

jackyzha0 (approved):

router/result.ts
bradymadden97: i'll try it! it might.

router/result.ts
bradymadden97: I'm not going to want to type that haha but if you'd prefer it will do

router/result.ts
bradymadden97: tru let me see there's probably a way.

router/result.ts
bradymadden97: oh i see this is slightly different behavior. mine was leaving the Ok/payload wrappers. Yours is just doing the payload.

router/result.ts
bradymadden97: added tests

PR #111, opened 2024-05-09
[protocol v2] RFC PROTOCOL V2

masad-frost:
## Why
Currently half-closes initiated by the server are not documented and sort of undefined behavior in the implementation. This is critical to adding `INVALID_REQUEST` errors.

## What changed

- Document `Error` type more explicitly
- Document `Result` type more explicitly
- Document `Control` type more explicitly
- Propose protocol errors
- Proposed changes for server initiated closes
  - These can either be normal closure or a protocol error closure
- Document reader and writer semantics

The changes should be somewhat backwards compatible, existing servers should be able to serve new clients with the exception of server initiated closes/protocol errors, I don't anticipate us needing to do a crazy migration to handle these changes.

## Versioning

- [ ] Breaking protocol change
- [ ] Breaking ts/js API change

PR #118, opened 2024-05-11
[protocol v2] RFC: custom streams

masad-frost:
## Why

`Pushable` from `it-pushable` is a convenient stream-looking abstraction to use, but it doesn't have the right affordances for consumers or the constraints we want in order to achieve good close semantics as proposed in #111. Web streams would have been nice to use, but they have a similar problem.

## What changed

Built some stream interfaces to be used by consumers in server handlers or on the client side. Things are documented inside. The e2e tests and related are updated to give an idea of how things will look.

Decided to leave implementation out of this PR just to get comments on the idea.

### DIFF
--- a/router/streams.ts
+++ b/router/streams.ts
@@ -1,0 +1,153 @@
+interface MinimalAsyncIterator<T> {
+  next(): Promise<
+    | {
+        done: false;
+        value: T;
+      }
+    | {
+        done: true;
+        value: undefined;
+      }
+  >;
+}
+
+interface MinimalAsyncIterable<T> {
+  [Symbol.asyncIterator](): MinimalAsyncIterator<T>;
+}
+
+/**
+ * A `ReadStream` represents a stream of data.
+ *
+ * This stream is not closable by the reader, the reader must wait for the writer to close it.
+ *
+ * The stream can only be consumed once. All consumers apply a permanent lock on the stream.
+ *
+ * To avoid memory leaks, ensure the stream is fully drained when it is no longer needed.
+ */
+export interface ReadStream<T> {
+  /**
+   * `iter` consumes/locks the stream and returns an iterable that can
+   * be used to iterate over the stream.
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  iter(): MinimalAsyncIterable<T>;
+  /**
+   * `drain` consumes/locks the stream and discards the content.
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  drain(): undefined;
+  /**
+   * `asArray` consumes/locks the stream and returns a promise that resolves
+   * with an array of the stream's content when the stream is closed.
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  asArray(): Promise<Array<T>>;
+  /**
+   * `tee` splits the stream into two {@link ReadStream} instances that
+   * can be consumed independently. The original stream will be locked forever.
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  tee(): [ReadStream<T>, ReadStream<T>];
+  /**
+   * `waitForClose` returns a promise that resolves when the stream is closed.
+   */
+  waitForClose(): Promise<void>;
+  /**
+   * `isClosed` returns true if the stream was closed by the writer.
+   */
+  isClosed(): boolean;
+  /**
+   * `isLocked` returns true if the stream is being consumed (aka locked).
+   */
+  isLocked(): boolean;
+  /**
+   * `breakConsumer` interrupts the consumer of the stream, causing it to throw an {@link BreakConsumerError}.
+   * The stream will remain locked forever and draining. If there is no consumer, this method is synonymous
+   * with {@link drain}.
+   */
+  breakConsumer(): undefined;
+  /**
+   * `breakConsumerUnsafe` same as {@link breakConsumer}, but doesn't throw an error. The reason
+   * it is unsafe is that the consumer will not be notified that the stream was interrupted, but
+   * they can manually check by calling {@link isBroken}.
+   */
+  breakConsumerUnsafe(): undefined;
+  /**
+   * `isBroken` returns true if the consumer was interrupted using {@link breakConsumer}
+   * or {@link breakConsumerUnsafe}.
+   */
+  isBroken(): boolean;
+  /**
+   * `requestClose` sends a request to the writer to close the stream, and resolves when the stream
+   * is fully closed. The stream can still receive more data after the request is sent.
+   */
+  requestClose(): undefined;
+  /**
+   * `isCloseRequested` checks if the reader has requested to close the stream.
+   */
+  isCloseRequested(): boolean;
+}
+
+/**
+ * `BreakConsumerError` is an error that is thrown when the consumer is interrupted.
+ */
+export class BreakConsumerError extends Error {
+  constructor() {
+    super('Consumer was interrupted');
+    this.name = 'BreakConsumerError';
+
+    if ('captureStackTrace' in Error) {
+      Error.captureStackTrace(this, BreakConsumerError);
+    }
+  }
+}
+
+/**
+ * A `WriteStream` is a streams that can be written to.
+ */
+export interface WriteStream<T> {
+  /**
+   * `write` writes a value to the stream. An error is thrown if writing to a closed stream.
+   */
+  write(value: T): undefined;
+  /**
+   * `close` signals the closure of the write stream, informing the reader that the stream is complete.
+   * Calling close multiple times has no effect.
+   */
+  close(): undefined;
+  /**
+   * `isCloseRequested` checks if the reader has requested to close the stream.
+   */
+  isCloseRequested(): boolean;
+  /**
+   * `waitForCloseRequest` returns a promise that resolves when the reader requests
+   * to close the stream.
+   */
+  waitForCloseRequest(): Promise<void>;
+  /**
+   * `waitForClose` returns a promise that resolves when the stream is closed
+   */
+  waitForClose(): Promise<void>;
+  /**
+   * `isClosed` returns true if the stream was closed by the writer.
+   */
+  isClosed(): boolean;
+}
+
+/**
+ * A `ReadWriteStream` combines a Readable and Writable stream.
+ */
+export interface ReadWriteStream<TRead, TWrite> {
+  /**
+   * The reader side of the stream.
+   */
+  reader: ReadStream<TRead>;
+  /**
+   * The writer side of the stream.
+   */
+  writer: WriteStream<TWrite>;
+}


masad-frost:
Still some things to uncover

router/streams.ts
```diff
@@ -0,0 +1,152 @@
...
+  /**
+   * `isLocked` returns true if the stream is being consumed (aka locked).
+   */
+  isLocked(): boolean;
+  /**
+   * `breakConsumer` interrupts the consumer of the stream, causing it to throw an error,
+   * the stream will remain locked forever and discards the content. If there is no consumer,
+   * this method is synonymous with `sink`.
+   */
+  breakConsumer(): void;
```
masad-frost, line 68: I'm thinking about making it so that this doesn't throw and stream consumers can check `isBroken` if they want to check that they got the full data.

The alternative here is to make the a `broken` state part of the iteration/values of the stream, but making everyone checks this for every iteration seems excessive.

masad-frost: I added `breakConsumerUnsafe`

masad-frost:
This feels bit over engineered and yet another layer to river :cry:, but I can't think of a better way to provide ergonomic streams that force devs into pit of success.

bradymadden97:
Maybe it would help to lay out a specific scenario where the pushable API breaks down, then we can figure out how to solve it? That part isn't clear to me, so it's hard to see how a new API helps.

masad-frost:
> Maybe it would help to lay out a specific scenario where the pushable API breaks down, then we can figure out how to solve it? That part isn't clear to me, so it's hard to see how a new API helps.

The pushable API exposed 2 methods (push and end), async iterator, and no properties. Specific scenarios are already expressed in the protocol PR.

For the writer, exposing an async iterator makes no sense and is confusing for it to be part of the API. There's also no way to get a signal that he reader is interested in closing. You can also call end multiple times, and you can start pushing to it again at any point, there's no way to tell if the steam has been closed. We'll also want to add abort it's semantics at some point. Prior to #111, we don't even need a pushable, just a simple write function.

On the reader side, it's a lot worse. First, there's no locking mechanism, so you could have 2 iterations fighting for the reads and you wouldn't know. The push method is exposed, and it makes no sense for a reader to have this. Getting a signal if the stream ended requires you to iterate over the whole pushable. You can call end on the pushable, I guess to signal that you're no longer interested, but the API isn't self explanatory, and end has unintended side effects like ending your iterator even though the stream didn't really end, it's just a request to close, so you lose your ability to know when the stream actually closes entirely. Prior to #111, we just need an asynciterable, not a pushable.

router/streams.ts
```diff
@@ -0,0 +1,159 @@
...
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  iter(): MinimalAsyncIterable<T>;
+  /**
+   * `sink` consumes/locks the stream and discards the content.
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  sink(): void;
```
masad-frost, line 40: probably drain

__tests__/fixtures/services.ts
```diff
@@ -157,19 +161,20 @@ export const FallibleServiceSchema = ServiceSchema.define({
...
-          returnStream.push(
+          writer.write(
             Err({
               code: STREAM_ERROR,
               message: 'field throwResult was set to true',
             }),
           );
         } else {
-          returnStream.push(Ok({ response: msg }));
+          writer.write(Ok({ response: msg }));
```
Monkatraz, line 177: I feel like using "write" terminology for the output stream is good, but "read" for the input stream feels not as intuitive. I feel like I want to use the word "observe" here more, and in-fact I think a true observable (so it's allowed to complete, or emit errors) fits really well here, and has a nice intuitive understanding. It may also be more flexible.

In the end this doesn't _really_ matter and using "read" is intuitive if you're used to e.g. buffers from Node, so it's fine. I also think it's better than what came before, I feel like this API is more naturally explanatory.

router/streams.ts
```diff
@@ -0,0 +1,159 @@
...
+export interface ReadWriteStream<TRead, TWrite> {
+  /**
+   * The reader side of the stream.
+   */
+  reader: CloseableReadStream<TRead>;
+  /**
+   * The writer side of the stream.
+   */
+  writer: WriteStream<TWrite>;
+}
```
Monkatraz, line 157: There are invariants between `reader` and `writer` here, right? Like, if the `writer` closes the stream, `reader` is going to be closed, right? It feels awkward to have to go and do something like `stream.reader.isClosed()` to check, because it feels like you should also have to check `stream.writer.isClosed()`, even if you don't. So maybe for `ReadWriteStream` it should lift those methods up?

router/streams.ts
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * `asArray` consumes/locks the stream and returns a promise that resolves
+   * with an array of the stream's content when the stream is closed.
+   */
+  asArray(): Promise<Array<T>>;
+  /**
+   * `tee` splits the stream into two {@link ReadStream} instances that
+   * can be consumed independently. The original stream will be locked forever.
+   */
+  tee(): [ReadStream<T>, ReadStream<T>];
```
Monkatraz, line 50: What is this for?

__tests__/fixtures/services.ts
```diff
@@ -45,10 +45,14 @@ const testServiceProcedures = TestServiceScaffold.procedures({
     init: Type.Object({ prefix: Type.String() }),
     input: EchoRequest,
     output: EchoResponse,
-    async handler(_ctx, init, msgStream, returnStream) {
-      for await (const { ignore, msg } of msgStream) {
+    async handler(_ctx, init, { reader, writer }) {
+      void writer.waitForCloseRequest().then(() => {
+        writer.close();
+      });
```
Monkatraz, line 51: I feel like this might be a common enough pattern that you could make a method for it or something - like `writer.setCloseWhenRequested(true)`

Monkatraz: Although elsewhere you do more than just close the stream when this promise resolves...

I wonder if adding teardowns would be a good here, like:
```ts
// called when the stream is closed
// maybe should just be onClose
writer.addTeardown(dispose)
```

Although I suppose you could just do this:
```ts
writer.waitForClose().then(dispose)
```

router/streams.ts
masad-frost: If you want to read in 2 different places, it allows you to create separate writers.

router/streams.ts
masad-frost: No they're completely independent. Each end (server or client) has its own reader for writes coming from the other side, and its own writer to write to the other side's reader.

masad-frost: Let me update the protocol doc to make the reader and writer semantics more clear, I'll ping you when that's done. But this is basically modeling after how most bidirectional streams work, e.g. http/2

Monkatraz: Okay, that makes sense. Also, maybe this should just be called `BidirectionalStream`?

__tests__/fixtures/services.ts
Monkatraz: After reading:
> No they're completely independent. Each end (server or client) has its own reader for writes coming from the other side, and its own writer to write to the other side's reader.

I feel more comfortable with this, although I feel like when I read this code I want to ask "reading from what?" and "writing to what?". If you are viewing streams as iterators, maybe you'd get confused that you're reading from the stream you're writing to, or you're writing to the stream you read from, which isn't the case, as these are independent streams.

__tests__/fixtures/services.ts
masad-frost: Yeah I wasn't sure which is more common. Wrapping callbacks in a promise tends to be messier than calling `.then` on a promise

router/streams.ts
masad-frost: Yeah, that works.

masad-frost: https://github.com/replit/river/pull/111/commits/1ed4b40a4cefbbafd6950d6f66317924aedcd995

__tests__/fixtures/services.ts
masad-frost: Maybe on the server `inputReader` and `outputWriter`, and on the client `outputReader` and `inputWriter`, since `input` and `output` are part of the api nomenclature

router/streams.ts
```diff
@@ -0,0 +1,153 @@
...
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  iter(): MinimalAsyncIterable<T>;
+  /**
+   * `drain` consumes/locks the stream and discards the content.
+   *
+   * Consuming a locked stream will throw an error.
+   */
+  drain(): undefined;
```
jackyzha0, line 40: hm my rust brain thought is that asArray should do what drain does here: https://doc.rust-lang.org/std/vec/struct.Drain.html

router/streams.ts
```diff
@@ -0,0 +1,153 @@
...
+   * The stream will remain locked forever and draining. If there is no consumer, this method is synonymous
+   * with {@link drain}.
+   */
+  breakConsumer(): undefined;
+  /**
+   * `breakConsumerUnsafe` same as {@link breakConsumer}, but doesn't throw an error. The reason
+   * it is unsafe is that the consumer will not be notified that the stream was interrupted, but
+   * they can manually check by calling {@link isBroken}.
+   */
+  breakConsumerUnsafe(): undefined;
```
jackyzha0, line 78: when would you need this

router/streams.ts
```diff
@@ -0,0 +1,153 @@
...
+  isBroken(): boolean;
+  /**
+   * `requestClose` sends a request to the writer to close the stream, and resolves when the stream
+   * is fully closed. The stream can still receive more data after the request is sent.
+   */
+  requestClose(): undefined;
+  /**
+   * `isCloseRequested` checks if the reader has requested to close the stream.
+   */
+  isCloseRequested(): boolean;
```
jackyzha0, line 92: feels like these two should only be present on the bidi pair not on just the pure ends

jackyzha0: but i guess in the actual usage you only ever have access to just one of the ends

router/streams.ts
masad-frost: Interesting idea. I imagine in most cases where you want this you can iterate and call a callback when you have a buffer of some length.

router/streams.ts
masad-frost: My thought was that you might want to interrupt an iterator/consumer without causing them to throw for some reason. I ended up not incorporating this in the final design, `drain` interrupts now.

router/streams.ts
masad-frost: In other words, a reader will always have a writer, it's not related to bidi streams

PR #120, opened 2024-05-12
[o11y] Add opentelemetry tracing

lhchavez:
## Why

We want to be able to know what pid2 is doing, and one good way of achieving that is through tracing.

This change is better reviewed with [whitespace turned off :sunglasses:](https://github.com/replit/river/pull/120/files?diff=split&w=1).

part of https://linear.app/replit/issue/WS-3107/pid2-spans

## What changed

This change adds opentelemetry tracing to river so that we know what's going on.

https://app.datadoghq.com/apm/traces?query=service%3Apid2%20-%40component%3Adns&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&colsTraces=service%2Cresource_name%2C%40duration%2C%40_span.count%2C%40_duration.by_service&fromUser=true&graphType=flamegraph&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=all&spanViewType=metadata&target-span=a&tq_query_translation_version=v0&traceQuery=a&view=spans&viz=stream&start=1715557200000&end=1715557260000&paused=true

## Versioning

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@replit/river",
   "description": "It's like tRPC but... with JSON Schema Support, duplex streaming and support for service multiplexing. Transport agnostic!",
-  "version": "0.19.3",
+  "version": "0.20.0",
   "type": "module",
   "exports": {
     ".": {
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
   },
   "peerDependencies": {
     "@sinclair/typebox": "~0.32.8",
+    "@opentelemetry/api": "^1.7.0",
     "isomorphic-ws": "^5.0.0",
     "ws": "^8.13.0"
   },
--- a/router/server.ts
+++ b/router/server.ts
@@ -1,3 +1,10 @@
+import {
+  context,
+  propagation,
+  type Span,
+  SpanKind,
+  SpanStatusCode,
+} from '@opentelemetry/api';
 import { Static } from '@sinclair/typebox';
 import { ServerTransport } from '../transport/transport';
 import { AnyProcedure, PayloadType } from './procedures';
@@ -22,6 +29,7 @@ import {
   ServiceContextWithTransportInfo,
 } from './context';
 import { log } from '../logging/log';
+import tracer from '../tracing';
 import { Value } from '@sinclair/typebox/value';
 import {
   Err,
@@ -181,6 +189,11 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
       return;
     }
 
+    let activeContext = context.active();
+    if (message.tracing) {
+      activeContext = propagation.extract(activeContext, message.tracing);
+    }
+
     const service = this.services[message.serviceName];
     const serviceContext = this.getContext(service, message.serviceName);
     if (!(message.procedureName in service.procedures)) {
@@ -247,12 +260,18 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
             }
           })();
 
-    const errorHandler = (err: unknown) => {
+    const errorHandler = (err: unknown, span: Span) => {
       const errorMsg = coerceErrorString(err);
       log?.error(
         `procedure ${message.serviceName}.${message.procedureName} threw an uncaught error: ${errorMsg}`,
         session.loggingMetadata,
       );
+      if (err instanceof Error) {
+        span.recordException(err);
+      } else {
+        span.recordException(errorMsg);
+      }
+      span.setStatus({ code: SpanStatusCode.ERROR });
 
       outgoing.push(
         Err({
@@ -292,15 +311,34 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
             return;
           }
 
-          try {
-            const outputMessage = await procedure.handler(
-              serviceContextWithTransportInfo,
-              inputMessage.value,
-            );
-            outgoing.push(outputMessage);
-          } catch (err) {
-            errorHandler(err);
-          }
+          await tracer.startActiveSpan(
+            `${message.serviceName}.${message.procedureName}`,
+            {
+              attributes: {
+                component: 'river',
+                'river.method.kind': 'rpc',
+                'river.method.service': message.serviceName,
+                'river.method.name': message.procedureName,
+                'river.streamId': message.streamId,
+                'span.kind': 'server',
+              },
+              kind: SpanKind.SERVER,
+            },
+            activeContext,
+            async (span: Span) => {
+              try {
+                const outputMessage = await procedure.handler(
+                  serviceContextWithTransportInfo,
+                  inputMessage.value,
+                );
+                outgoing.push(outputMessage);
+              } catch (err) {
+                errorHandler(err, span);
+              } finally {
+                span.end();
+              }
+            },
+          );
         })();
         break;
       case 'stream':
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
   ],
   "dependencies": {
     "@msgpack/msgpack": "^3.0.0-beta2",
+    "@opentelemetry/api": "^1.7.0",
```
jackyzha0, line 52: should this be a peer dependency? otherwise the global registration won't work properly

router/server.ts
```diff
@@ -282,15 +295,39 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
-          try {
-            const outputMessage = await procedure.handler(
-              serviceContextWithTransportInfo,
-              inputMessage.value,
-            );
-            outgoing.push(outputMessage);
-          } catch (err) {
-            errorHandler(err);
-          }
+          await tracer.startActiveSpan(
```
jackyzha0, line 298: why not 

```ts
const span = tracer.startSpan("do operation");
// do something
span.end();
```

i guess my q is whether `startActiveSpan` gets us anything here because this really affects indentation level/blame which i want to avoid

package.json
lhchavez: prolly, please help with the fix since i'm not that well-versed in JS tooling.

router/server.ts
lhchavez: it does :( the machinery here traverses the stack to know what is the currently active context. the implication is that if there were internal spans in any of the service definitions, they _need_ this extra indentation.

let's bite the bullet, we need this.

package.json
lhchavez: ah nvm found it https://nodejs.org/en/blog/npm/peer-dependencies

router/server.ts
jackyzha0: for future readers: https://www.honeycomb.io/blog/startspan-versus-startactivespan-opentelemetry

jackyzha0 (changes requested):
i wonder if we can get a nicer wrapper API around tracer.startActiveSpan that does all the span stuff for us in a way that auto-ends the spans via a function decorator (seems like all the traces we care about are at the fn level)

this would help us avoid the indentation problem and make this a lot nicer to read

jackyzha0:
seems like they have an 'official' library way to register instrumentations that we should do:

1. example implementation https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L65
2. example usage https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-http#usage

lhchavez:
> seems like they have an 'official' library way to register instrumentations that we should do:
> 
> 1. base pkg https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation
> 2. example implementation https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/opentelemetry-instrumentation-http/src/http.ts#L65
> 3. example usage https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages/opentelemetry-instrumentation-http#usage

yeah, that's the recommended way to do it _if_ you don't have access to the code that you're instrumenting (because it's a lot more work involved, and a bit more jank). we do!

router/server.ts
```diff
@@ -282,15 +295,39 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
+            `${message.serviceName}.${message.procedureName}`,
+            {
+              attributes: {
+                component: 'river',
+                'river.method.kind': 'rpc',
+                'river.method.service': message.serviceName,
+                'river.method.name': message.procedureName,
+                'span.kind': 'server',
+              },
+              kind: SpanKind.SERVER,
```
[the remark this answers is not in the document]
lhchavez: ah because just using `kind:` here (the otel way) doesn't show this in the datadog context, which is nice.

router/server.ts
```diff
@@ -282,15 +295,39 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
+            activeContext,
+            async (span: Span) => {
+              try {
+                const outputMessage = await procedure.handler(
+                  serviceContextWithTransportInfo,
+                  inputMessage.value,
+                );
+                outgoing.push(outputMessage);
+              } catch (err) {
+                if (err instanceof Error) {
```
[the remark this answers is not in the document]
lhchavez: why not!

transport/transport.ts
```diff
@@ -625,89 +633,146 @@ export abstract class ClientTransport<
...
-      if (backoffMs > 0) {
-        sleep = new Promise((resolve) => setTimeout(resolve, backoffMs));
+    return tracer.startActiveSpan('connect', async (span: Span) => {
+      try {
+        await this.connectAttempt(to);
+      } catch (e: unknown) {
+        if (e instanceof Error) {
+          span.recordException(e);
+        } else {
+          span.recordException(String(e));
```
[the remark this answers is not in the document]
lhchavez: awww yiss!

transport/transport.ts
```diff
@@ -625,89 +633,146 @@ export abstract class ClientTransport<
...
-      log?.info(`attempting connection to ${to} (${backoffMs}ms backoff)`, {
-        clientId: this.clientId,
-        connectedTo: to,
-      });
-      this.retryBudget.consumeBudget(to);
-      reconnectPromise = sleep
-        .then(() => {
+  private async connectAttempt(
+    to: TransportClientId,
+    attempt = 0,
```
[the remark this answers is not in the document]
lhchavez: ah rats, forgot to add it to the context.

transport/transport.ts
```diff
@@ -717,8 +782,15 @@ export abstract class ClientTransport<
   }
 
   protected sendHandshake(to: TransportClientId, conn: ConnType) {
+    const tracing = { traceparent: '', tracestate: '' };
+    propagation.inject(context.active(), tracing);
```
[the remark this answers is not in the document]
lhchavez: /shrug i think it's because the OG implementation is Java. and you know how Java does things.

router/server.ts
```diff
@@ -301,25 +338,87 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
+                  const dispose = await procedure.handler(
+                    serviceContextWithTransportInfo,
+                    initMessage.value,
+                    incoming,
+                    outgoing,
+                  );
+
+                  if (dispose) {
+                    disposables.push(dispose);
+                    disposables.push(() => span.end());
```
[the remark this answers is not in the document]
lhchavez: hmm we can go with the same subspan for now. if service writers are doing beefy things in their dispose, they can add a subspan (i guess)

tracing/index.ts
```diff
@@ -0,0 +1,4 @@
+import { trace } from '@opentelemetry/api';
+
+const tracer = trace.getTracer('river');
```
masad-frost, line 3:
```suggestion
const tracer = trace.getTracer('river', RIVER_VERSION);
```

maybe?

lhchavez: oooh yes!

lhchavez:
@masad-frost @jackyzha0 if there are no more comments, mind approving so that we can get better o11y in production?

jackyzha0 (approved):
have a follow-up pr with some cleanup/code quality in a bit but lgtm for functionality

router/server.ts
```diff
@@ -339,19 +433,42 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
+            `${message.serviceName}.${message.procedureName}`,
+            {
+              attributes: {
+                component: 'river',
+                'river.method.kind': 'subscription',
+                'river.method.service': message.serviceName,
+                'river.method.name': message.procedureName,
+                'river.streamId': message.streamId,
+                'span.kind': 'server',
+              },
```
masad-frost, line 446: wonder if we move attributes to a `tracing` type so we don't miss updating one and not others.

router/server.ts
```diff
@@ -181,6 +189,11 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
       return;
     }
 
+    let activeContext = context.active();
+    if (message.tracing) {
+      activeContext = propagation.extract(activeContext, message.tracing);
+    }
```
masad-frost, line 195: I wonder if we should use [AsyncLocalStorageContextManager](https://www.npmjs.com/package/@opentelemetry/context-async-hooks) because with async stacks "activeSpan" will be all kinds of messed up. 

Seems like it's marked a stable in node 20 and bun
https://nodejs.org/docs/latest-v20.x/api/async_context.html#class-asynclocalstorage
https://bun.sh/docs/runtime/nodejs-apis#node-async-hooks

That said, I remember AsyncLocalStorage has major performance issues, not sure if tracer incurs those

router/server.ts
```diff
@@ -311,25 +349,81 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
+            {
+              attributes: {
+                component: 'river',
+                'river.method.kind': 'stream',
+                'river.method.service': message.serviceName,
+                'river.method.name': message.procedureName,
+                'river.streamId': message.streamId,
+                'span.kind': 'server',
+              },
+              kind: SpanKind.SERVER,
```
masad-frost, line 402: wonder if it's best to move these to a function to ensure we're consistent across the board. e.g. `startActiveSpan(getProcedureSpanArgs(message, kind))`

masad-frost (approved):
Let's see how this fairs without async context.

router/server.ts
masad-frost: I guess jacky will can take care of this

router/server.ts
masad-frost: jacky can follow up in his pr

router/server.ts
jackyzha0: https://www.npmjs.com/package/@opentelemetry/context-zone we might be able to use this?

masad-frost: I've used zone.js with Angular, and it had even bigger footguns wrt performance.

IIRC it also makes the package harder to consume due to it patching a lot of internals in the runtime, and last I looked at it, it requires weird UMD setup.

masad-frost: we can stick with this for now, if we notice our traces aren't getting nested properly, we can try AsyncLocalStorage and benchmark it.

masad-frost: For posterity: chatted offline, we're gonna go with managing context manually!

lhchavez: cool, with this convo resolved, i'm going to merge to start getting traces :rocket:

PR #123, opened 2024-05-14
Add overloading to bindLogger

masad-frost:
## Why

`const log: Logger = bindLogger(logfn, 'info')` returns a type that's `Logger | undefined`

## What changed

- Add overloading so that we can get instance of loggers easily.
- Remove logging level from `Logger` interface

## Versioning


masad-frost:
Maybe the better thing to do here is to expose `unbindLogger()`

jackyzha0 (approved):

PR #125, opened 2024-05-14
Emit sourcemaps

masad-frost:
## Why

We need this for good stack traces

## What changed

- Change tsup to emit sourcemaps

## Versioning


lhchavez (approved):
:ez: one-line fix

jackyzha0 (approved):

PR #136, opened 2024-05-21
Client uses ReadStream and WriteStream

masad-frost:
## Why

More protocolv2 work #127 

## What changed

In this PR it's a simple replacement of `pushable` interface with stream interfaces, on the client-side. No protocol or API changes otherwise. Streams API is a superset of `pushable`, many of the stream methods aren't implemented correctly/don't make a lot of sense without further changes, but this is a good step.

## Versioning

### DIFF
--- a/__tests__/bandwidth.bench.ts
+++ b/__tests__/bandwidth.bench.ts
@@ -52,13 +52,13 @@ describe('bandwidth', async () => {
       { time: BENCH_DURATION },
     );
 
-    const [input, output] = await client.test.echo.stream();
+    const [input, reader] = await client.test.echo.stream();
     bench(
       `${name} -- stream`,
       async () => {
         input.push({ msg: nanoid(), ignore: false });
-        const result = await output.next();
-        assert(result.value.ok);
+        const result = await reader[Symbol.asyncIterator]().next();
+        assert(result.value?.ok);
       },
       { time: BENCH_DURATION },
     );
--- a/__tests__/bandwidth.bench.ts
+++ b/__tests__/bandwidth.bench.ts
@@ -52,12 +52,12 @@ describe('bandwidth', async () => {
       { time: BENCH_DURATION },
     );
 
-    const [input, reader] = await client.test.echo.stream();
+    const [inputWriter, outputReader] = await client.test.echo.stream();
     bench(
       `${name} -- stream`,
       async () => {
-        input.push({ msg: nanoid(), ignore: false });
-        const result = await reader[Symbol.asyncIterator]().next();
+        inputWriter.write({ msg: nanoid(), ignore: false });
+        const result = await outputReader[Symbol.asyncIterator]().next();
         assert(result.value?.ok);
       },
       { time: BENCH_DURATION },
--- a/__tests__/cleanup.test.ts
+++ b/__tests__/cleanup.test.ts
@@ -182,9 +182,10 @@ describe.each(testMatrix())(
         clientTransport.eventDispatcher.numberOfListeners('message');
 
       // start procedure
-      const [input, outputReader, close] = await client.test.echo.stream();
-      input.push({ msg: '1', ignore: false, end: undefined });
-      input.push({ msg: '2', ignore: false, end: true });
+      const [inputWriter, outputReader, close] =
+        await client.test.echo.stream();
+      inputWriter.write({ msg: '1', ignore: false, end: undefined });
+      inputWriter.write({ msg: '2', ignore: false, end: true });
 
       const outputIterator = getIteratorFromStream(outputReader);
       const result1 = await iterNext(outputIterator);
@@ -193,7 +194,7 @@ describe.each(testMatrix())(
 
       // ensure we only have one stream despite pushing multiple messages.
       await waitFor(() => expect(server.streams.size).toEqual(1));
-      input.end();
+      inputWriter.close();
       // ensure we no longer have any streams since the input was closed.
       await waitFor(() => expect(server.streams.size).toEqual(0));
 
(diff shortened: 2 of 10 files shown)


router/client.ts
```diff
@@ -477,24 +489,31 @@ function handleSubscribe(
         payload: input,
         controlFlags: ControlFlags.StreamOpenBit,
       });
-
       let healthyClose = true;
 
       // transport -> output
-      const outputStream = pushable({ objectMode: true });
+      const readStreamRequestCloseNotImplemented = () => void 0;
```
Monkatraz, line 495: why `void 0` instead of `undefined`? or just `() => {}`?

router/client.ts
```diff
@@ -395,15 +381,31 @@ function handleStream(
...
-        }
+        },
+        () => {
+          // after closing input stream, send a close message to the server
+          if (!healthyClose) return;
+          transport.sendCloseStream(serverId, streamId);
+          span.setStatus({ code: SpanStatusCode.OK });
+        },
+      );
+      const readStreamRequestCloseNotImplemented = () => void 0;
```
Monkatraz, line 392: ditto

util/testHelpers.ts
```diff
@@ -74,12 +75,27 @@ export function createLocalWebSocketClient(port: number) {
...
+  next(): Promise<
+    | {
+        done: false;
+        value: T;
+      }
+    | {
+        done: true;
+        value: undefined;
+      }
+  >;
```
Monkatraz, line 97: maybe this should be its own type lol, seems kind of useful

__tests__/cleanup.test.ts
```diff
@@ -364,10 +368,11 @@ describe.each(testMatrix())(
       });
 
       // start a stream
-      const [input, output] = await client.test.echo.stream();
-      input.push({ msg: '1', ignore: false });
+      const [inputWriter, outputReader] = await client.test.echo.stream();
```
Monkatraz, line 371: I like you naming these `inputWriter` and `outputReader` (and on the server, `inputReader` and `outputWriter`), not too verbose but feels intuitive

util/testHelpers.ts
```diff
@@ -74,12 +75,27 @@ export function createLocalWebSocketClient(port: number) {
   return sock;
 }
 
+export function getIteratorFromStream<T>(readStream: ReadStream<T>) {
+  return readStream[Symbol.asyncIterator]();
+}
```
Monkatraz, line 80: I'm still kind of surprised there isn't something like `Iterator.from(foo)` in JS

Monkatraz (approved):
lgtm! just nitpicks or observations

router/client.ts
masad-frost: we have strict linting that screams at you when you have empty functions lol, `undefined` works.

__tests__/cleanup.test.ts
masad-frost: I'd like to update type annotations to have those names, but maybe we can do that in a follow up since we reference `'input'` and `'output'` in a bunch of places

router/client.ts
Monkatraz: blegh understandable then lol

util/testHelpers.ts
masad-frost: I had it be part of the interfaces in `streams.ts`, but then just inlined it, we'll see if it comes up again.

__tests__/bandwidth.bench.ts
```diff
@@ -52,13 +52,13 @@ describe('bandwidth', async () => {
...
-    const [input, output] = await client.test.echo.stream();
+    const [inputWriter, outputReader] = await client.test.echo.stream();
     bench(
       `${name} -- stream`,
       async () => {
-        input.push({ msg: nanoid(), ignore: false });
-        const result = await output.next();
-        assert(result.value.ok);
+        inputWriter.write({ msg: nanoid(), ignore: false });
+        const result = await outputReader[Symbol.asyncIterator]().next();
```
jackyzha0, line 60: does a bare `await outputReader().next();` not work?

__tests__/cleanup.test.ts
```diff
@@ -182,25 +182,27 @@ describe.each(testMatrix())(
...
-      const [input, output, close] = await client.test.echo.stream();
-      input.push({ msg: '1', ignore: false, end: undefined });
-      input.push({ msg: '2', ignore: false, end: true });
+      const [inputWriter, outputReader, close] =
+        await client.test.echo.stream();
+      inputWriter.write({ msg: '1', ignore: false, end: undefined });
+      inputWriter.write({ msg: '2', ignore: false, end: true });
 
-      const result1 = await iterNext(output);
+      const outputIterator = getIteratorFromStream(outputReader);
```
jackyzha0, line 190: also when to use `outputReader[Symbol.asyncIterator]` vs `getIteratorFromStream(outputReader)`

util/testHelpers.ts
jackyzha0: seems like we should be able to just use the stream as an async iterator https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/asyncIterator without this helper?

masad-frost: the for-await-of function automatically calls this, if you want to call `next` directly, you have to grab the iterator like this

__tests__/bandwidth.bench.ts
masad-frost: no, you need to grab the iterator

__tests__/cleanup.test.ts
masad-frost: it's just a helper, we can either remove `getIteratorFromStream` or i'll make sure we're using it everywhere.

PR #137, opened 2024-05-21
Server uses ReadStream and WriteStream #136 

masad-frost:
Exactly the same as #136 but for the server

### DIFF
--- a/router/server.ts
+++ b/router/server.ts
@@ -13,8 +13,6 @@ import {
   InstantiatedServiceSchemaMap,
   AnyServiceSchemaMap,
 } from './services';
-import { pushable } from 'it-pushable';
-import type { Pushable } from 'it-pushable';
 import {
   ControlMessagePayloadSchema,
   OpaqueTransportMessage,
@@ -41,6 +39,7 @@ import {
 import { EventMap } from '../transport/events';
 import { Connection } from '../transport/session';
 import { coerceErrorString } from '../util/stringify';
+import { ReadStreamImpl, WriteStreamImpl } from './streams';
 
 /**
  * Represents a server with a set of services. Use {@link createServer} to create it.
@@ -56,10 +55,9 @@ interface ProcStream {
   id: string;
   serviceName: string;
   procedureName: string;
-  incoming: Pushable<PayloadType>;
-  outgoing: Pushable<Result<Static<PayloadType>, Static<RiverError>>>;
+  incoming: ReadStreamImpl<PayloadType>;
+  outgoing: WriteStreamImpl<Result<Static<PayloadType>, Static<RiverError>>>;
   promises: {
-    outputHandler: Promise<unknown>;
     inputHandler: Promise<unknown>;
   };
 }
@@ -116,7 +114,7 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
     }
 
     let procStream = this.streamMap.get(message.streamId);
-    const isInitMessage = !procStream;
+    const isFirstMessage = !procStream;
 
     // create a proc stream if it doesnt exist
     procStream ||= this.createNewProcStream(message);
@@ -125,7 +123,7 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
       return;
     }
 
-    await this.pushToStream(procStream, message, isInitMessage);
+    await this.pushToStream(procStream, message, isFirstMessage);
   };
 
   // cleanup streams on session close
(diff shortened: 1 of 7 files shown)


router/server.ts
```diff
@@ -580,25 +602,22 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
+    if (!isFirstMessage || !hasInit) {
+      // Init message is consumed during stream instantiation
+      if (Value.Check(procedure.input, message.payload)) {
+        procStream.incoming.pushValue(message.payload as PayloadType);
+      } else if (!Value.Check(ControlMessagePayloadSchema, message.payload)) {
+        log?.error(
+          `procedure ${serviceName}.${procedureName} received invalid payload`,
+          { clientId: this.transport.clientId, fullTransportMessage: message },
+        );
+      }
```
masad-frost, line 620: I don't like this approach (or any of the changes i made to this file for that matter.) There's going to have to be some refactors for us to implement invalid requests, half-close properly, and close requests, so I'm gonna stick to keeping changes to here as minimal as possible.

router/server.ts
```diff
@@ -580,25 +602,22 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
-    } else if (Value.Check(procedure.input, message.payload)) {
-      procStream.incoming.push(message.payload as PayloadType);
-    } else if (!Value.Check(ControlMessagePayloadSchema, message.payload)) {
-      log?.error(
-        `procedure ${serviceName}.${procedureName} received invalid payload`,
-        { clientId: this.transport.clientId, fullTransportMessage: message },
-      );
+    const hasInit = 'init' in procedure;
+
+    if (!isFirstMessage || !hasInit) {
```
masad-frost, line 611: in theory this shouldn't work for `rpc`, but it does :shrug:

router/server.ts
```diff
@@ -634,10 +653,11 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
     }
 
     // end the streams and wait for the handlers to finish
-    stream.incoming.end();
+    if (!stream.incoming.isClosed()) {
+      stream.incoming.triggerClose();
+    }
```
masad-frost, line 658: should probably call `drain` here. Also don't need to be checking `isClosed`, but it's ok for now

router/server.ts
masad-frost: We probably just end up pushing to the stream twice.

router/server.ts
```diff
@@ -326,12 +302,24 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
             },
             activeContext,
             async (span: Span) => {
+              if (!Value.Check(procedure.input, message.payload)) {
```
jackyzha0, line 305: shouldn't the first (and only message) on a subscription to the server be init? im a little confused, is this + #136 rewriting to also do the thing we talked about where all streams are mandatory stream init message + potentially more body messages instead of the current state of the world where it is optional special init message + maybe more body messages

router/server.ts
jackyzha0: what do we end up pushing 🤔

router/server.ts
jackyzha0: yeah this is pretty hard to parse in some points, it feels inconsistent how and when we check init vs payload at least within the diff

util/testHelpers.ts
```diff
@@ -205,27 +226,34 @@ export function asClientStream<
...
   extendedContext?: Omit<ServiceContext, 'state'>,
   session: Session<Connection> = dummySession(),
-) {
-  const input = pushable<Static<I>>({ objectMode: true });
-  const output = pushable<Result<Static<O>, Static<E>>>({
-    objectMode: true,
-  });
+): [WriteStream<Static<I>>, ReadStream<ProcedureResult<O, E>>] {
+  const inputPipe = createPipe<Static<I>>();
+  const outputPipe = createPipe<ProcedureResult<O, E>>();
```
jackyzha0, line 231: ooo this is nice

jackyzha0 (approved):

router/server.ts
masad-frost: This is the RPC handler btw, first message don't go into the stream.

Yeah if we have mandatory inits the code will be much cleaner

router/server.ts
```diff
@@ -448,10 +444,22 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
```
masad-frost, line 447: yeah this should check init :/

masad-frost: or wait, subscriptions don't have `init`, only `input`.

router/server.ts
masad-frost: subscriptions and rpc currently are not `init`. I'll update the protocol doc then readapt the changes

PR #140, opened 2024-05-22
improve tracing

jackyzha0:
## Why + What Changed

- we'd like better abstractions around spans/context so we don't clutter all our functions with callbacks
- we'd also like spans for more river-native concepts like sessions + connections and more events around the lifecycle

this pr, like the last otel pr, is best reviewed with whitespace changes off :)

## Versioning

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -613,9 +613,9 @@
       }
     },
     "node_modules/@opentelemetry/api": {
-      "version": "1.7.0",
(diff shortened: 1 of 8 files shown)


masad-frost:
Cries in rebase #127

masad-frost (approved):
hadouken'd

image


Seems legit, i don't have a good intuition on how spans should be structured, but this seems to be 1:1 with what he previously (given that we didn't have child spans to make use of startActiveSpan).

Issue #119, opened 2024-05-11
Improve tsconfig.json


masad-frost:
The tsconfig is pretty stock, we should fix it. `lib` should be `[]`, and then there are some of the strict checks like `noUncheckedIndexedAccess` that we should enable. It'll lead to some pedantic type checking, but it's probably fine for this library.

PR #133, opened 2024-05-21
all my homies hate isomorphic-ws

jackyzha0:
## Why

- [isomorphic-ws](https://github.com/heineiuo/isomorphic-ws) is jank and basically does a dumb switch on platform without trying to do any sort of compat behaviour between ws and the Brower's WebSocket
- this is a very thin comptability shim around both WebSockets that satisfies a common subset of the two APIs

## What changed

- refer to code in https://github.com/jackyzha0/agnostic-ws

## Versioning

- [ ] Breaking protocol change
- [ ] Breaking ts/js API change

PR #141, opened 2024-05-23
Go back to accepting a WebSocket for WebSocketClientTransport

masad-frost:
## Why

So `agnostic-ws` from #133, while less janky, caused similar issues with `isomorphic-ws` when it came to module resolution and global WebSocket, this showed up when using jest and it was spooky.

## What changed

- Once again we accept a WebSocket getter in the `WebSocketClientTransport` constructor
  - To enable this, we created an interface `WSLike` which uses (ugly) generics to make it so that we can pass the web WebSocket (types specified in `lib.dom.d.ts`)  or one of the Node websocket modules that are almost compatible with the web WebSocket.
  - `WebSocketConnection` constructor accepts `WSLike` as well.
- Unnested `createNewOutgoingConnection` in `WebSocketClientTransport` by using async-await more 
- Revived `createLocalWebSocketClient` test helper
- Updated `WebSocketConnection` error listener to emit close code and reason, this was done by using `onclose` on the websocket, and checking if `onerror` was called previously, the calling the `error` listener set by the connection user.
- Denoised `tsconfig` by removing comments
- Updated `lib` in `tsconfig` to be explicitly `lib: []`. Previously, it was set to the default which include `"dom"` and might've caused issues if for example we referenced `WebSocket` or other web apis.

## Versioning


.replit
```diff
@@ -1,4 +1,4 @@
-run = "npm run test:ui"
+run = "npm run test"
```
masad-frost, line 1: this started a VNC since it launches browser lol

transport/impls/ws/wslike.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * A websocket-like interface that has all we need, this matches
+ * "lib.dom.d.ts" and npm's "ws" websocket interfaces.
+ *
+ * This makes things a little weird within the ws module but it's
+ * the only way to avoid casting and type errors.
+ */
```
masad-frost, line 7:
```suggestion
 * This makes things a little ugly with generics within the ws transport module but
 * it's the only way to avoid casting and type errors.
 */
```

.replit
```diff
@@ -12,6 +12,10 @@ channel = "stable-23_05"
 localPort = 3000
 externalPort = 80
 
+[[ports]]
+localPort = 24678
+externalPort = 3000
+
```
jackyzha0, line 18: extra?

package.json
```diff
@@ -50,13 +50,12 @@
   "dependencies": {
     "@msgpack/msgpack": "^3.0.0-beta2",
     "it-pushable": "^3.2.3",
-    "nanoid": "^4.0.2"
+    "nanoid": "^4.0.2",
+    "ws": "^8.17.0"
```
jackyzha0, line 54: no longer peer for ws? we should try to keep the ws lib version the same right? or maybe doesn't matter now that we have the interface

transport/impls/ws/client.ts
```diff
@@ -1,25 +1,46 @@
...
+export class WebSocketClientTransport<
+  CloseEvent extends { code: number; reason: string; wasClean: boolean } = {
+    code: number;
+    reason: string;
+    wasClean: boolean;
+  },
+  MessageEvent extends { data: unknown } = { data: unknown },
+  ErrorEvent extends object = object,
+  OpenEvent extends object = object,
+  BinaryType extends string = string,
```
jackyzha0, line 24: do these need to be specified at the transport level?

transport/impls/ws/client.ts
```diff
@@ -29,70 +50,55 @@ export class WebSocketClientTransport extends ClientTransport<WebSocketConnectio
...
+    });
 
-      this.handleConnection(conn, to);
-      return conn;
-    } else {
-      throw new Error(wsRes.err);
-    }
+    this.handleConnection(conn, to);
+    return conn;
   }
```
jackyzha0, line 103: ty for cleaning this up lol

transport/impls/ws/connection.ts
```diff
@@ -19,15 +57,15 @@ export class WebSocketConnection extends Connection {
   }
 
   addCloseListener(cb: () => void): void {
-    this.ws.onclose = cb;
+    this.closecb = cb;
```
jackyzha0, line 60:
```suggestion
    this.closeCb = cb;
```

for casing consistency

masad-frost: I thought this is Go :P

package.json
masad-frost: Given the way we use it in WebSocket server transport, it doesn't make sense not to pull it in directly.

transport/impls/ws/connection.ts
masad-frost: i wonder why event listeners like onclose or onbeforeunload in web are all lower case lol

jackyzha0: yeah actually have no idea lol just thought about this too (no strong pref tbh)

.replit
masad-frost: @joe-thom i wish there was a way to disable port detection / auto editing my .replit

jackyzha0 (approved):

jackyzha0:
Also closes #119

PR #144, opened 2024-05-23
[nix] Add flake.nix with dev shell

cbrewster:
## Why

This makes it easy for developers who use Nix/NixOS to develop river.

## What changed

- Add `flake.nix` with dev shell including node & typescript-language-server
- Add `.envrc` for `direnv` support, which also adds node binaries to the `PATH` for convenience

## Versioning

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,25 @@
+{
+  description = "It's like tRPC but... with JSON Schema Support, duplex streaming and support for service multiplexing. Transport agnostic!";
+
+  inputs.nixpkgs.url = "github:nixos/nixpkgs";
+
+  outputs = { self, nixpkgs }:
+  let
+    mkDevShell = system:
+    let
+      pkgs = nixpkgs.legacyPackages.${system};
+    in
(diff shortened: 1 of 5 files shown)


cbrewster:

* main:
  * **PR #144** 👈

flake.nix
```diff
@@ -0,0 +1,25 @@
...
+  outputs = { self, nixpkgs }:
+  let
+    mkDevShell = system:
+    let
+      pkgs = nixpkgs.legacyPackages.${system};
+    in
+    pkgs.mkShell {
+      nativeBuildInputs = with pkgs; [
+        nodejs
+        nodePackages.typescript-language-server
```
jackyzha0, line 15: we should add eslint language server too

https://github.com/NixOS/nixpkgs/pull/239650

jackyzha0 (approved):

flake.nix
cbrewster: We get eslint from node and its automatically put in the path from:
```
PATH_add ./node_modules/.bin
```

PR #147, opened 2024-05-25
make sure we dont assign conn until handshake is complete

jackyzha0:
## Why

`getOrCreateSession` assigns the connection to the session if the connection is passed in. The way we used it before meant that we could pass the connection to the session _before_ the handshake completed which mean that you could have

1. start connection
2. session is created and handshake is initiated
3. client gets session notif and synchronously tries to send a message
4. handshake is finished

to the server, it would get the client message before the handshake message which violates a protocol assumption

## What changed

literally just dont assign the connection to the session until the handshake is complete

## Versioning

### DIFF
--- a/transport/transport.test.ts
+++ b/transport/transport.test.ts
@@ -97,6 +97,34 @@ describe.each(testMatrix())(
       }
     });
 
+    test('sending right after session event should not cause invalid handshake', async () => {
+      const clientTransport = getClientTransport('client');
+      const protocolError = vi.fn();
+      clientTransport.addEventListener('protocolError', protocolError);
+      const serverTransport = getServerTransport();
+
+      const msg = createDummyTransportMessage();
+      const msgPromise = waitForMessage(serverTransport);
+      const sendHandle = (evt: EventMap['sessionStatus']) => {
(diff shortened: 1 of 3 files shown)


Monkatraz (approved):

transport/transport.ts
```diff
@@ -824,7 +824,9 @@ export abstract class ClientTransport<
       }
     }
 
-    const { session } = this.getOrCreateSession(to, conn);
+    // dont pass conn here as we dont want the session to start using the conn
+    // until we have finished the handshake
+    const { session } = this.getOrCreateSession(to);
```
masad-frost, line 829: Feels like this method is inherently messy, feels like we should have 2 separate methods, you're either creating a session or you expect one to be there, no?

transport/transport.test.ts
```diff
@@ -97,6 +97,34 @@ describe.each(testMatrix())(
       }
     });
 
+    test('sending right after session event should not cause invalid handshake', async () => {
```
masad-frost, line 100: Noice

masad-frost (approved):
Good enough to unblock, but i think we can do better.

PR #149, opened 2024-05-27
don't log the full transport message in logs, add validation info

jackyzha0:
## Why

this caused issues in production with large payload responses (e.g. buffers containing long files)

## What changed

- strip payload and tracing info from logs
- log parsing errors where appropriate

## Versioning

### DIFF
--- a/logging/log.ts
+++ b/logging/log.ts
@@ -1,5 +1,5 @@
+import { ValueError } from '@sinclair/typebox/value';
 import { OpaqueTransportMessage } from '../transport';
-import { PartialTransportMessage } from '../transport/message';
 
 const LoggingLevels = {
   debug: -1,
@@ -20,17 +20,26 @@ export type Logger = {
 
 export type Tags = 'invariant-violation';
 
-export type MessageMetadata = Record<string, unknown> &
-  Partial<{
-    protocolVersion: string;
-    clientId: string;
-    connectedTo: string;
-    sessionId: string;
-    connId: string;
-    fullTransportMessage: OpaqueTransportMessage;
-    partialTransportMessage: Partial<PartialTransportMessage>;
-    tags: Array<Tags>;
-  }>;
+function cleanMessage<T extends Partial<OpaqueTransportMessage>>(
+  msg: T,
+): Omit<T, 'payload' | 'tracing'> {
+  const { payload, tracing, ...rest } = msg;
+  return rest;
+}
+
+export type MessageMetadata = Partial<{
+  protocolVersion: string;
+  clientId: string;
+  connectedTo: string;
+  sessionId: string;
+  connId: string;
+  transportMessage: Omit<
+    Partial<OpaqueTransportMessage>,
+    'payload' | 'tracing'
+  >;
+  validationErrors: Array<ValueError>;
+  tags: Array<Tags>;
+}>;
 
 class BaseLogger implements Logger {
   minLevel: LoggingLevel;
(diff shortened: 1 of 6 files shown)


logging/log.ts
```diff
@@ -86,19 +95,51 @@ export const jsonLogger: LogFn = (msg, ctx, level) => {
...
+    {
+      get(_target, prop: LoggingLevel) {
+        return (msg: string, metadata?: MessageMetadata) => {
+          // skip cloning object if metadata has no transportMessage
+          if (!metadata?.transportMessage) {
+            log[prop](msg, metadata);
+            return;
+          }
+
+          // clone metadata and clean transportMessage
```
masad-frost, line 114: Do we need to copy the whole metadata? I think we can clone only the transport message

logging/log.ts
```diff
@@ -86,19 +95,51 @@ export const jsonLogger: LogFn = (msg, ctx, level) => {
   console.log(JSON.stringify({ msg, ctx, level }));
 };
 
+let _log: Logger | undefined = undefined;
+
+// log proxy to clean transport payloads
 export let log: Logger | undefined = undefined;
+const createLogProxy = (log: Logger) =>
```
masad-frost, line 102: We need to wrap 4 known functions, why don't we do that instead of using a proxy. Proxies are great when there's variable fields. Maybe im missing something though.

logging/log.ts
```diff
@@ -86,19 +95,51 @@ export const jsonLogger: LogFn = (msg, ctx, level) => {
...
+  new Proxy(
+    {},
+    {
+      get(_target, prop: LoggingLevel) {
```
masad-frost, line 106: There might be other functions in a passed logger, right?

logging/log.ts
```diff
@@ -86,19 +95,51 @@ export const jsonLogger: LogFn = (msg, ctx, level) => {
...
+          // skip cloning object if metadata has no transportMessage
+          if (!metadata?.transportMessage) {
+            log[prop](msg, metadata);
+            return;
+          }
+
+          // clone metadata and clean transportMessage
+          log[prop](msg, {
+            ...metadata,
+            transportMessage: cleanMessage(metadata.transportMessage ?? {}),
```
masad-frost, line 117: We already check for the transport message above this shouldn't be needed

logging/log.ts
```diff
@@ -86,19 +95,51 @@ export const jsonLogger: LogFn = (msg, ctx, level) => {
...
+let _log: Logger | undefined = undefined;
+
+// log proxy to clean transport payloads
 export let log: Logger | undefined = undefined;
+const createLogProxy = (log: Logger) =>
+  new Proxy(
+    {},
+    {
+      get(_target, prop: LoggingLevel) {
+        return (msg: string, metadata?: MessageMetadata) => {
```
masad-frost, line 107: Maybe we shouldn't strip the data for debug logs?

masad-frost: I've definitely made use of it.

transport/transport.ts
```diff
@@ -322,6 +322,9 @@ export abstract class Transport<ConnType extends Connection> {
     if (!Value.Check(OpaqueTransportMessageSchema, parsedMsg)) {
       log?.error(`received invalid msg: ${JSON.stringify(parsedMsg)}`, {
         clientId: this.clientId,
+        validationErrors: [
+          ...Value.Errors(OpaqueTransportMessageSchema, parsedMsg),
```
masad-frost, line 326: Nice

masad-frost (changes requested):
Maybe we should consider doing this at the passed logger layer. In any case, left some ideas.

logging/log.ts
jackyzha0: ya fair point, i had this before i did the wrapper approach so its an artifact of a previous approach

masad-frost (approved):
Approve in case this is costing us and needs to be patched quick

logging/log.ts
jackyzha0: i think i need this spread to override `transportMessage` without affecting the actual message itself

logging/log.ts
```diff
@@ -86,19 +100,36 @@ export const jsonLogger: LogFn = (msg, ctx, level) => {
   console.log(JSON.stringify({ msg, ctx, level }));
 };
 
+let _log: Logger | undefined = undefined;
```
masad-frost, line 103: what's this for?

jackyzha0: `_log` is the actual underlying Logger, the `log` is the log proxy that the rest of the library calls into with the full info and it cleans it before calling the underlying `_log`

we just do `Logger | undefined` in the proxy so that we can avoid constructing the objects when a logger is not binded

PR #151, opened 2024-05-28
pass previous metadata to validate

jackyzha0:
## Why + What changed

- session no longer contains previous metadata info so passing it to validate doesn't make sense
- let's just pass the metadata directly (this is the only thing in session we check in web anyways)
- added a test for this

## Versioning

### DIFF
--- a/transport/message.ts
+++ b/transport/message.ts
@@ -1,6 +1,5 @@
 import { Type, TSchema, Static } from '@sinclair/typebox';
 import { nanoid } from 'nanoid';
-import { Connection, Session } from './session';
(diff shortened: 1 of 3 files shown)


masad-frost (approved):
I forgot why we added in the first place, but looks good.

jackyzha0:
> I forgot why we added in the first place, but looks good.

i think specifically to handle cases where you may reconnect with a token that says you are someone else (though i think this risk is mostly a non-issue)

in any case, this helps us disambiguate connect vs reconnect in the validation

PR #153, opened 2024-05-28
[protocolv2] always expect init, input sometimes (docs only)

masad-frost:
## Why

Currently we expect:
- 0 init messages for `rpc` and `subscription`, and 0-1 init messages for `upload` and `stream`
- 1-n input messages for `upload` and `stream`, and 0 input for 

This works but if we did:
- 1 init message for all
- 0-n input messages (always 0 for `rpc` and `subscription`)

We'd have a shape of streams that is a little more consistent, making implementations simpler, although practically it's not much simpler since it's papered over in the libraries.

## What changed

Updated documentation for `init` and `input`.


masad-frost:
In practice this doesn't affect the implementations since wire messages don't differentiate between `init` and `input`.

PR #154, opened 2024-05-29
remove unnecessary generic `MetadataSchema`

jackyzha0:
## Why

- transports dont need the type of the schema `MetadataSchema` drilled throughout
- construction of extra handshake data is internal
- everything else is just `ParsedMetadata` which is exposed as an interface that can be overridden

## What changed

- remove unnecessary generic `MetadataSchema`
- export `ParsedMetadata` so we can properly override it

## Versioning

### DIFF
--- a/transport/message.ts
+++ b/transport/message.ts
@@ -16,7 +16,9 @@ export const enum ControlFlags {
   StreamClosedBit = 0b0100,
 }
 
-export interface ClientHandshakeOptions<MetadataSchema extends TSchema> {
+export interface ClientHandshakeOptions<
+  MetadataSchema extends TSchema = TSchema,
+> {
   /**
    * Schema for the metadata that the client sends to the server
    * during the handshake.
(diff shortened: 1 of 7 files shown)


transport/message.ts
```diff
@@ -16,7 +16,9 @@ export const enum ControlFlags {
   StreamClosedBit = 0b0100,
 }
 
-export interface ClientHandshakeOptions<MetadataSchema extends TSchema> {
+export interface ClientHandshakeOptions<
+  MetadataSchema extends TSchema = TSchema,
+> {
   /**
    * Schema for the metadata that the client sends to the server
    * during the handshake.
```
masad-frost, line 24: maybe we shouldn't have user pass this, and we swap `MetadataSchema` with `unknown` in construct? I don't think it provides a lot of value

masad-frost (approved):

transport/message.ts
jackyzha0: we have the generic here to constrain construct to return the static type of the schema

PR #159, opened 2024-05-30
[protocolv2] Init always exists

masad-frost:
## Why

Follow up to #153 

## What changed

- `Init` is on every type of rpc
- `Input` only for `upload` and `stream`

Mostly deleted code after I did the replacement, hide whitespace to see that it's mostly deleting code https://github.com/replit/river/pull/159/files?w=1

Some tests don't work because it ended up affecting close semantics. Tests should start working once we implement half-closes, which I'm almost done with.

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -95,9 +95,9 @@ interface BaseError {
 The `Result` type MUST conform to:
 
 ```ts
-type Result<T, E extends BaseError> =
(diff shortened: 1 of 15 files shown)


masad-frost:
Jacky please take a look when you get a chance, I'm gonna start merging some of this stuff into the base branch, I'll address review in follow ups.

PR #160, opened 2024-05-31
[protocolv2] swap ReadStream's waitForClose promise with callback style onClose

masad-frost:
## Why

We discussed this in some of the earlier PRs for ReadStream. After using it, I decided that it's a lot cleaner to do a callback and wrap with a promise as needed instead of using `.then` as a callback when needed

## What changed

Title says it all

### DIFF
--- a/router/streams.ts
+++ b/router/streams.ts
@@ -43,10 +43,11 @@ export interface ReadStream<T> {
    */
   isLocked(): boolean;
   /**
-   * `waitForClose` returns a promise that resolves when the stream is closed,
-   * does not send a close request.
+   * `onClose` registers a callback that will be called when the stream
+   * is closed. Returns a function that can be used to unregister the
+   * listener.
    */
-  waitForClose(): Promise<undefined>;
+  onClose(cb: () => void): () => void;
   /**
    * `isClosed` returns true if the stream was closed by the writer.
    *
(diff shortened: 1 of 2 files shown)


router/streams.ts
```diff
@@ -136,13 +137,9 @@ export class ReadStreamImpl<T> implements ReadStream<T> {
...
   /**
-   * Used to for `waitForClose` and `requestClose`.
+   * A list of listeners that will be called when the stream is closed.
    */
-  private closePromise: Promise<undefined>;
-  /**
-   * Resolves closePromise
-   */
-  private resolveClosePromise: () => void = () => undefined;
+  private onCloseListeners: Set<() => void>;
```
Monkatraz, line 142: There is a utility exported from `@replit/observable` called `ListenerList` that would be helpful here. You can use it like this:
```ts
const list = new ListenerList()
const dispose = list.add(() => {})
list.notify()
```
or, if you want to notify with arguments:
```ts
const list = new ListenerList<(msg: string, int: number) => void>()
const dispose = list.add((msg, int) => {})
list.notify('foo', 1)
// notify inherits the type parameter's arguments
```

Monkatraz (approved):
lgtm, just one comment about maybe using `ListenerList`

router/streams.ts
masad-frost: Looks like we haven't published it

masad-frost: this seems simple enough, it's not really worth a utility imo

masad-frost: if we end up with a lot of listeners, we can swap them out.

Monkatraz: oh right I totally forgot that this is in river lmao

PR #161, opened 2024-05-31
🧹  Use nix cache for eslint language server on replit

bradymadden97:
## Why

we don't have to pollute the package.json to have eslint in river on replit. we can use the nix cache for the package

## What changed

eslint still works 
image

## Versioning


jackyzha0 (approved):

masad-frost (approved):
ty

PR #162, opened 2024-05-31
[protocolv2] Half-close client

masad-frost:
## Why

Starting to implement half-close semantics, this targets the client side. Please read Protocol.md in #127 to understand what's going on.

## What changed

- Some API changes
  - No more exposing "cleanup"/"close" functions, clients can only close inputWriter streams or request the outputWriter on the server to close.
  - procedure calls shape changes
    - `stream` returns an `[inputWriter, outputReader]` tuple
    - `upload` returns an `[inputWriter, finalizerFn]` tuple, where `finalizerFn` returns `Promise<OutputResult>`
    - `subscription` returns `outputReader`
    - `rpc` returns `Promise<OutputResult>`
- I extracted procedure function types to make it easy to type check our proxy's methods
- Replaced the various proc handlers with one main one (similar to how `createProcStream` works on the server), and then construct the return interface.
  - Bit of an aside: This makes sense since the underlying streams for all procedures should be the same. Even things like `rpc` sends `CloseBit` and `OpenBit` as part of `init` is an optional thing, and would work for other types of procs if we want. In theory, with `rpc` you can send init with an `OpenBit` only and then send a `CloseControl` message as a follow. In a more extreme world, your `stream` can send `CloseBit` and `OpenBit` as part of `init` which instantly closes the `input` (client-to-server) stream. We might want to update `Protocol.md` to indicate that all types of procedures work the same way, and things like `rpc` sending open and close in one message is just an optimization.
  - Subscription now sends a `CloseBit` with `init`, since we don't need the `input` stream.

### DIFF
--- a/router/client.ts
+++ b/router/client.ts
@@ -14,11 +14,17 @@ import {
   ControlFlags,
   TransportClientId,
   isStreamClose,
-  PartialTransportMessage,
+  ControlMessageCloseSchema,
 } from '../transport/message';
 import { Static } from '@sinclair/typebox';
 import { nanoid } from 'nanoid';
-import { Err, Result, UNEXPECTED_DISCONNECT } from './result';
+import {
+  AnyResultSchema,
+  Err,
+  Result,
+  RiverError,
+  UNEXPECTED_DISCONNECT,
+} from './result';
 import { EventMap } from '../transport/events';
 import { Connection } from '../transport/session';
 import { log } from '../logging/log';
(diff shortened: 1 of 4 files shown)


masad-frost:
Test will fail miserably as I deferred fixing them to the server implementation

router/result.ts
```diff
@@ -104,27 +120,25 @@ export type Output<
     ? Procedure extends object & { rpc: infer RpcHandler extends Fn }
       ? Awaited<ReturnType<RpcHandler>>
       : Procedure extends object & { upload: infer UploadHandler extends Fn }
-      ? Awaited<ReturnType<UploadHandler>> extends [
+      ? ReturnType<UploadHandler> extends [
           infer __UploadInputMessage,
-          infer UploadOutputMessage,
+          (...args: never) => Promise<infer UploadOutputMessage>,
```
jackyzha0, line 125: o damn nice

router/result.ts
```diff
@@ -40,6 +40,22 @@ export const RiverUncaughtSchema = Type.Object({
...
 
+export const AnyResultSchema = Type.Union([
+  Type.Object({
+    ok: Type.Literal(false),
+    payload: Type.Object({
+      code: Type.String(),
+      message: Type.String(),
+      extras: Type.Optional(Type.Unknown()),
+    }),
+  }),
```
jackyzha0, line 51:
```suggestion
  Type.Object({
    ok: Type.Literal(false),
    payload: RiverErrorSchema,
  }),
```

router/result.ts
```diff
@@ -104,27 +120,25 @@ export type Output<
...
           infer __UploadInputMessage,
-          infer UploadOutputMessage,
+          (...args: never) => Promise<infer UploadOutputMessage>,
         ]
-        ? Awaited<UploadOutputMessage>
+        ? UploadOutputMessage
         : never
       : Procedure extends object & { stream: infer StreamHandler extends Fn }
-      ? Awaited<ReturnType<StreamHandler>> extends [
+      ? ReturnType<StreamHandler> extends [
```
jackyzha0, line 130: still weird that some procedure constructors are async and others aren't but maybe thats fine 🤷

jackyzha0 (approved):

router/client.ts
```diff
@@ -206,334 +244,209 @@ export function createClient<ServiceSchemaMap extends AnyServiceSchemaMap>(
...
+      log?.error('Received message after stream is closed', {
+        clientId: transport.clientId,
+        transportMessage: msg,
+      });
+
+      return;
+    }
+
+    if (!Value.Check(ControlMessageCloseSchema, msg.payload)) {
+      if (Value.Check(AnyResultSchema, msg.payload)) {
```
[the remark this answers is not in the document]
masad-frost: yeah great idea.

router/result.ts
masad-frost: I tried to make `RiverErrorSchema` a real type and it was a pain in the butt. I got something working but it was so gnarly. I can pull that out and we can look into it later.

router/result.ts
masad-frost: It's just Rpc that returns a promise

router/client.ts
```diff
@@ -30,8 +36,62 @@ import {
   WriteStream,
   WriteStreamImpl,
 } from './streams';
-
-export type AsyncIter<T> = AsyncGenerator<T, T>;
+import { Value } from '@sinclair/typebox/value';
+import { PayloadType, ValidProcType } from './procedures';
+
+type RPCFn<
```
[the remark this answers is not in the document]
masad-frost: HTML or Html, URL or Url? :shrug:

lol i'll change it on the base branch

router/result.ts
masad-frost: I put it in a Linear task

PR #163, opened 2024-05-31
[protocolv2] Half-close server

masad-frost:
## Why

Continuation for #162, server implementation

## What changed

Just implementing half close for server. I'll annotate some changes in comments

## Versioning

### DIFF
--- a/__tests__/typescript-stress.test.ts
+++ b/__tests__/typescript-stress.test.ts
@@ -268,12 +268,8 @@ describe('Output<> type', () => {
     }
 
     // Then
-    void client.test.stream
-      .stream({})
-      .then(([_in, outputReader, _close]) =>
-        iterNext(getIteratorFromStream(outputReader)),
-      )
-      .then(acceptOutput);
+    const [, outputReader] = client.test.stream.stream({});
+    void iterNext(getIteratorFromStream(outputReader)).then(acceptOutput);
     expect(client).toBeTruthy();
   });
 
@@ -286,12 +282,9 @@ describe('Output<> type', () => {
     }
 
     // Then
-    void client.test.subscription
-      .subscribe({ n: 1 })
-      .then(([outputReader, _close]) =>
-        iterNext(getIteratorFromStream(outputReader)),
-      )
-      .then(acceptOutput);
+    const outputReader = client.test.subscription.subscribe({ n: 1 });
+    void iterNext(getIteratorFromStream(outputReader)).then(acceptOutput);
+
     expect(client).toBeTruthy();
   });
 
@@ -302,10 +295,9 @@ describe('Output<> type', () => {
     }
 
     // Then
-    void client.test.upload
-      .upload({})
-      .then(([_input, result]) => result)
-      .then(acceptOutput);
+    const [, finalize] = client.test.upload.upload({});
+    void finalize().then(acceptOutput);
+
     expect(client).toBeTruthy();
   });
 });
--- a/router/server.ts
+++ b/router/server.ts
@@ -46,15 +46,17 @@ export interface Server<Services extends AnyServiceSchemaMap> {
   close(): Promise<void>;
 }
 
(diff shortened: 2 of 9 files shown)


router/server.ts
```diff
@@ -520,18 +499,27 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
       return;
     }
 
     // end the streams and wait for the handlers to finish
-    if (!stream.incoming.isClosed()) {
-      stream.incoming.triggerClose();
+    if (!stream.inputReader.isClosed()) {
+      // Drain will make sure any read leads to an error instead of looking like
+      // the stream cleanly closed.
+      stream.inputReader.drain();
```
masad-frost, line 512: We decided to switch that to not throw, and instead make the `inputReader` also use `Result`. Will do that in a follow up.

router/server.ts
```diff
@@ -249,12 +278,16 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
-      );
+
+      if (!outputWriter.isClosed()) {
+        outputWriter.write(
+          Err({
+            code: UNCAUGHT_ERROR,
+            message: errorMsg,
+          } satisfies Static<typeof RiverUncaughtSchema>),
+        );
+        outputWriter.close();
```
masad-frost, line 289: Error handling is basically not implemented yet

image


image

Will do that in the future

router/server.ts
```diff
@@ -520,18 +499,27 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
+    if (!stream.inputReader.isClosed()) {
+      // Drain will make sure any read leads to an error instead of looking like
+      // the stream cleanly closed.
+      stream.inputReader.drain();
+      stream.inputReader.triggerClose();
+    }
+    // We wait for the handler because we want to ensure that we have all the disposables
+    void stream.inputHandlerPromise.then((maybeDispose) => {
+      maybeDispose?.();
+    });
```
masad-frost, line 518: I don't like the way this works. I think in general `cleanupStream` will be phased out. Will revisit with error handling and might refactor more of the server

router/server.ts
```diff
@@ -494,7 +473,7 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
     }
 
     if (isStreamClose(message.controlFlags)) {
-      await this.cleanupStream(message.streamId);
+      procStream.inputReader.triggerClose();
 
       const streamsFromThisClient = this.clientStreams.get(message.from);
       if (streamsFromThisClient) {
```
masad-frost, line 479: this whole code path is wrong. Will fix up in a follow, we shouldn't be immediately removing the stream as the outputWriter can still be oepn

__tests__/typescript-stress.test.ts
```diff
@@ -302,10 +295,9 @@ describe('Output<> type', () => {
     }
 
     // Then
-    void client.test.upload
-      .upload({})
-      .then(([_input, result]) => result)
-      .then(acceptOutput);
+    const [, finalize] = client.test.upload.upload({});
+    void finalize().then(acceptOutput);
```
jackyzha0, line 299: upload api is still weird to me, like most uploads dont seem like they need `Init` and I would expect some uniform `Input` and finalize would just be ending the Writable Stream and wait for some promise (so this matches the unary server return case of Rpc)

masad-frost: > like most uploads dont seem like they need Init 
Mm, I disagree, but don't have a lot of upload examples. In my mind, most of the time you wanna target a resource and you need an init to specify it. The only `upload` we have in the code right now is `fs.writeFile` which needs to specify the file path and the size of the file.

> I would expect some uniform Input and finalize would just be ending the Writable Stream and wait for some promise (so this matches the unary server return case of Rpc)

That's what the new API looks like.

masad-frost: (I merged, we can half the API shape discussion separately from this PR)

PR #164, opened 2024-05-31
[protocolv2] swap WriteStream's waitForCloseRequest promise with callback style onCloseRequest

masad-frost:
Exactly like #160 but for the write stream.

### DIFF
--- a/__tests__/streams.test.ts
+++ b/__tests__/streams.test.ts
@@ -361,7 +361,9 @@ describe('WriteStream unit', () => {
 
     expect(stream.isCloseRequested()).toBeFalsy();
 
-    const closeRequestP = stream.waitForCloseRequest();
+    const closeRequestP = new Promise<void>((resolve) =>
+      stream.onCloseRequest(resolve),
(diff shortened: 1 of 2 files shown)


__tests__/streams.test.ts
```diff
@@ -361,7 +361,9 @@ describe('WriteStream unit', () => {
 
     expect(stream.isCloseRequested()).toBeFalsy();
 
-    const closeRequestP = stream.waitForCloseRequest();
+    const closeRequestP = new Promise<void>((resolve) =>
```
jackyzha0, line 364: what if we had a `stream.promises.closeRequest`-style api kinda like fs

jackyzha0 (approved):

__tests__/streams.test.ts
masad-frost: Let's try to use this API and see if we have a lot of promise wrappers. I don't think we'll see a lot of of use for promises for this outside of contrived situations like tests.

PR #165, opened 2024-06-01
[protocolv2] Implement close requests

masad-frost:
## Why

We decided in protocolv2 that readers can requests writers to close (though writers might not comply). More details in Protocol.md in #127 

## What changed

When `reader.requestClose` is called, we send a close control message with `StreamCloseRequestBit`. The writer side's should then receive request and decide what to do with it.


transport/message.ts
```diff
@@ -10,9 +10,10 @@ import { PropagationContext } from '../tracing';
  * it will not be using the stream anymore.
  */
 export const enum ControlFlags {
-  AckBit = 0b0001,
-  StreamOpenBit = 0b0010,
-  StreamClosedBit = 0b0100,
+  AckBit = 0b00001,
+  StreamOpenBit = 0b00010,
+  StreamClosedBit = 0b00100,
+  StreamCloseRequestBit = 0b10000,
```
jackyzha0, line 16: maybe add comments for these i actually forget the distinction here

jackyzha0 (approved):

__tests__/cleanup.test.ts
```diff
@@ -183,17 +183,18 @@ describe.each(testMatrix())(
 
       // start procedure
       const [inputWriter, outputReader] = client.test.echo.stream({});
-      inputWriter.write({ msg: '1', ignore: false, end: undefined });
-      inputWriter.write({ msg: '2', ignore: false, end: true });
```
jackyzha0, line 187: this actually was helpful to test the presence of undefined fields in an object that had an optional fields but we could probably pull that test out elsewhere

masad-frost: Oh, yeah I'll move it to another test so no one makes the same mistake

PR #168, opened 2024-06-03
Adds OTEL tests, fixes some issues with propagation

ksh-chy:
## Why

There were some small issues with propagation due to how we associate context's with spans. When we create a span, we need to ensure that we use `trace.setSpan` to associate the span with a context. The `trace.startSpan` function accepts a context as well, but this is meant to be used to access the parent span, and the context is not updated by itself. 

## What changed

Add some dev dependencies for testing, updated functions in `tracing/index.ts` to follow the convention above

## Versioning


util/testHelpers.ts
```diff
@@ -24,6 +35,18 @@ import { Connection, Session, SessionOptions } from '../transport/session';
...
+  const provider = new BasicTracerProvider();
+  provider.addSpanProcessor(
+    new SimpleSpanProcessor(new InMemorySpanExporter()),
+  );
+  const contextManager = new StackContextManager();
+  contextManager.enable();
+  trace.setGlobalTracerProvider(provider);
+  context.setGlobalContextManager(contextManager);
+  propagation.setGlobalPropagator(new W3CTraceContextPropagator());
+});
```
jackyzha0, line 48: seems weird to do this at a global level, can we do this just for tracing tests?

jackyzha0 (approved):

util/testHelpers.ts
ksh-chy: Done!

tracing/tracing.test.ts
```diff
@@ -0,0 +1,75 @@
...
+  SimpleSpanProcessor,
+} from '@opentelemetry/sdk-trace-base';
+import { W3CTraceContextPropagator } from '@opentelemetry/core';
+import { StackContextManager } from '@opentelemetry/sdk-trace-web';
+
+import tracer from './index';
+import * as tracing from './index';
+import { OpaqueTransportMessage } from '../transport';
+
+beforeAll(() => {
```
jackyzha0, line 19: sorry last nit: does it still work if we move it inside the describe block?

ksh-chy: Yeah that seems to be fine, let me commit this change

PR #171, opened 2024-06-03
integrated tracing test

jackyzha0:
## Why

- we should test that spans get propagated correctly for sessions, etc. and that parent/child relationships work as expected

## What changed

- fix up some tests
- add integration matrix for session spans

## Versioning


masad-frost (approved):
nice ty

PR #174, opened 2024-06-05
Support diffing server schemas for API incompatibilities

cbrewster:
## Why

With upcoming versioning work, we want a way to be able to programmatically detect API compatibility breakages between different River schemas.

This adds a first pass at schema diffing which detects a number of incompatibilities:
- Removed services
- Removed procedures
- Procedure type changes
- Payload incompatibilities
    - Type changes
    - New required fields on client->server messages
    - Removed required fields on server->client messages
    - The check support nested objects
    - Array changes
    - Union changes
    - Intersection changes

## What changed

- Add `diffServerSchema` which diffs two server schemas and returns all detected API incompatibilies
- Includes a test suite checking for the incompatibilies mentioned above

## Versioning

### DIFF
--- a/router/diff.ts
+++ b/router/diff.ts
@@ -1,0 +1,220 @@
+import type {
+  SerializedServerSchema,
+  SerializedServiceSchema,
+  SerializedProcedureSchema,
+  PayloadType,
+} from '../router';
+
+export interface ServerBreakage {
+  serviceBreakages: Record<string, ServiceBreakage>;
+}
+
+export type ServiceBreakage =
+  | {
+      reason: 'removed';
+    }
+  | {
+      reason: 'modified';
+      procedureBreakages: Record<string, ProcedureBreakage>;
+    };
+
+export type ProcedureBreakage =
+  | {
+      reason: 'removed';
+    }
+  | { reason: 'type-changed'; oldType: string; newType: string }
+  | {
+      reason: 'modified';
+      input?: PayloadBreakage;
+      init?: PayloadBreakage;
+      output?: PayloadBreakage;
+    };
+
+export type PayloadBreakage =
+  | { reason: 'type-changed'; oldType: string; newType: string }
+  | { reason: 'new-required' }
+  | { reason: 'removed-required' }
+  | {
+      reason: 'field-breakage';
+      fieldBreakages: Record<string, PayloadBreakage>;
+    };
+
+export function diffServerSchema(
+  oldServer: SerializedServerSchema,
+  newServer: SerializedServerSchema,
+): ServerBreakage | null {
+  const allServices = new Set([
+    ...Object.keys(oldServer.services),
+    ...Object.keys(newServer.services),
+  ]);
+
+  const breakages: Record<string, ServiceBreakage> = {};
+  for (const serviceName of allServices) {
+    const oldService = oldServer.services[serviceName];
+    const newService = newServer.services[serviceName];
+
+    const breakage = diffService(oldService, newService);
+    if (breakage) {
+      breakages[serviceName] = breakage;
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


cbrewster:

* main:
  * **PR #174** 👈

router/diff.ts
```diff
@@ -0,0 +1,220 @@
...
+      return { reason: 'new-required' };
+    }
+    // New service, perfectly fine.
+    return null;
+  }
+
+  if (oldPayload.type !== newPayload.type) {
+    return {
+      reason: 'type-changed',
+      oldType: oldPayload.type,
```
cbrewster, line 187: Not sure what to do about these errors. Ideally I could've used `oldPayload[Kind]` here, but we are using `Type.Strict` on the serialized form to omit the compositing symbols, so `type` seems like the best option, but need to persuade Typescript that this is fine.

router/diff.ts
```diff
@@ -0,0 +1,220 @@
...
+    };
+
+export type PayloadBreakage =
+  | { reason: 'type-changed'; oldType: string; newType: string }
+  | { reason: 'new-required' }
+  | { reason: 'removed-required' }
+  | {
+      reason: 'field-breakage';
+      fieldBreakages: Record<string, PayloadBreakage>;
+    };
```
cbrewster, line 40: Not sure if this is the best way to structure the breakage, open to suggestions. Ultimately I want to wire this up to a CLI which provides a pretty printed version of the breakages with some explanations.

router/diff.ts
```diff
@@ -0,0 +1,220 @@
...
+    newProcedure.init,
+    'client',
+  );
+  const outputBreakage = diffPayload(
+    oldProcedure.output,
+    newProcedure.output,
+    'server',
+  );
+
+  if (inputBreakage || initBreakage || outputBreakage) {
```
cbrewster, line 139: Is the linter right? Using `??` here feels to weird...

bradymadden97: I think for your code they are functionally the same, but `||` will treat some other things like 0 as falsy in js so I think that's where the warning comes from. `??` is fine

image

router/diff.ts
masad-frost: yeah, that's fine, we can just ignore them.

We have some pedantic lint rules that are often useful, but sometimes annoying when you're doing something that's not super structured/well typed.

router/diff.ts
masad-frost: this seems fine, if we weren't using json schema i would've suggested using typebox's `Errors` helper.

masad-frost: nvm, derp

router/diff.ts
```diff
@@ -558,7 +565,6 @@ function diffJSONSchema(
         };
       }
     }
-    console.log('yo', newSchema);
```
masad-frost: lol

router/diff.ts
```diff
@@ -212,6 +212,13 @@ function diffRequired(
     return null;
   }
 
+  if (origin === 'client' && !oldRequired && newRequired) {
+    return { reason: 'new-required' };
+  }
+  if (origin === 'server' && oldRequired && !newRequired) {
+    return { reason: 'removed-required' };
+  }
```
masad-frost, line 220: ah good catch

cbrewster:
We might also want to support checking for forwards-compatibility. We may want to allow newer pid2 workspace clients to work with older versions of pid2 servers, this way we don't have to force upgrade pid2 so often. If we can check forwards compatibility we can add and enforce a _minimum server version_ that the client has and sends to the server during the handshake.

cbrewster:
Sounds like we're going to punt on forwards compat checking for now, we can add that later if we need it.

PR #175, opened 2024-06-06
[protocolv2] Add abort support, move protocol errors to abort status (docs only)

masad-frost:
## Why

`StreamClosedBit` should be reserved for normal closures where the writer decides to close gracefully, overloading it to transmit `ProtocolError`s which lead to abrupt closure can be confusing and problematic. Since we plan on adding abort/cancellation support, might as well go ahead and do this now while we're rolling out a breaking change 

## What changed

Added `StreamAbortBit` and moved protocol errors there. This also involves us adding abort support instead of clean closes.

## Versioning

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -146,7 +146,7 @@ The `controlFlags` property is a [bit field](https://en.wikipedia.org/wiki/Bit_f
 - The `StreamOpenBit` (`0b00010`) MUST be set for the first message of a new stream.
 - The `StreamClosedBit` (`0b00100`) MUST be set for the last message of a stream.
 - The `StreamCloseRequestBit` (`0b10000`) MUST be set for a message that is requesting the other side to close the stream.
-- Bits `0b01000` is reserved for future use and is currently unused.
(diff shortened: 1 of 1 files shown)


PROTOCOL.md
```diff
@@ -146,7 +146,7 @@ The `controlFlags` property is a [bit field](https://en.wikipedia.org/wiki/Bit_f
 - The `StreamOpenBit` (`0b00010`) MUST be set for the first message of a new stream.
 - The `StreamClosedBit` (`0b00100`) MUST be set for the last message of a stream.
 - The `StreamCloseRequestBit` (`0b10000`) MUST be set for a message that is requesting the other side to close the stream.
-- Bits `0b01000` is reserved for future use and is currently unused.
+- The `StreamAbortBit` (`0b01000`) MUST be set when a stream is to be abruptly closed due to cancellations or an internal error condition.
```
masad-frost, line 149: I'm considering swapping the bits for `StreamAbortBit` and `StreamClosedBit`. The old implementation of `StreamClosedBit` is actually closer to an abort, though the don't carry the right payload (`ControlClose` instead of `ProtocolError`), the implementations don't seem to peek the payload anyway just the control flag.

jackyzha0 (approved):

PR #177, opened 2024-06-09
[dev] Deflake the tests

lhchavez:
## Why

The tests were flaky. Really flaky. They didn't show up their flakiness because they were only run once, but any time they were repeated more than once, it was Flake City, USA.

## What changed

This change does quite a few things:
* Stops using vitest's global `onTestFinished` in favor of the per-test one: the former runs _asynchronously_, and also runs only once per test suite (not per test case). This means that all the cleanup and timer shenanigans are done concurrently with other tests (even if vitest only does file-level parallelism!), so that messes up the global state a lot.
  * in other words, it is completely legal for two copies of `cleanup.ts`' `testFinishesCleanly()` to run concurrently with each other. The first copy will call `vi.useFakeTimers(...)`, hit an `await` point, and then control passes to the second copy, which will hit the line where `vi.useRealTimers()` is reached. Then the first copy runs again and will continue running thinking everything is fine, try to advance a timer and explode.
* Tells vitest to run all hooks as stacks, and avoid file parallelism for extra correctness.

## Versioning

### DIFF
--- a/__tests__/cleanup.test.ts
+++ b/__tests__/cleanup.test.ts
@@ -1,12 +1,4 @@
-import {
-  afterAll,
-  assert,
-  describe,
-  expect,
-  test,
-  vi,
-  onTestFinished,
-} from 'vitest';
+import { afterAll, assert, describe, expect, test, vi } from 'vitest';
 import { iterNext } from '../util/testHelpers';
 import {
   SubscribableServiceSchema,
@@ -31,7 +23,9 @@ describe.each(testMatrix())(
       await transport.setup({ client: opts, server: opts });
     afterAll(cleanup);
 
-    test('closing a transport from the client cleans up connection on the server', async () => {
+    test('closing a transport from the client cleans up connection on the server', async ({
+      onTestFinished,
+    }) => {
       // setup
       const clientTransport = getClientTransport('client');
       const serverTransport = getServerTransport();
@@ -71,7 +65,9 @@ describe.each(testMatrix())(
       await ensureTransportBuffersAreEventuallyEmpty(serverTransport);
     });
 
-    test('closing a transport from the server cleans up connection on the client', async () => {
+    test('closing a transport from the server cleans up connection on the client', async ({
+      onTestFinished,
+    }) => {
       // setup
       const clientTransport = getClientTransport('client');
       const serverTransport = getServerTransport();
@@ -111,7 +107,7 @@ describe.each(testMatrix())(
       await ensureTransportBuffersAreEventuallyEmpty(serverTransport);
     });
 
-    test('rpc', async () => {
+    test('rpc', async ({ onTestFinished }) => {
       // setup
       const clientTransport = getClientTransport('client');
       const serverTransport = getServerTransport();
@@ -158,7 +154,7 @@ describe.each(testMatrix())(
       await ensureTransportBuffersAreEventuallyEmpty(serverTransport);
     });
 
-    test('stream', async () => {
+    test('stream', async ({ onTestFinished }) => {
       // setup
       const clientTransport = getClientTransport('client');
       const serverTransport = getServerTransport();
@@ -226,7 +222,7 @@ describe.each(testMatrix())(
       await ensureTransportBuffersAreEventuallyEmpty(serverTransport);
     });
 
-    test('subscription', async () => {
+    test('subscription', async ({ onTestFinished }) => {
       // setup
       const clientTransport = getClientTransport('client');
       const serverTransport = getServerTransport();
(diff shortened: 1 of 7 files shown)


jackyzha0 (commented):
talked in slack but i believe the `vi.useFakeTimers(...)` that we use in certain tests outside of the cleanup are problematic if run in parallel, we can try removing [file parallelism](https://vitest.dev/config/#fileparallelism) so that we know we are running exactly one test at a time

i also believe it is a bug that `onTestFinished` runs only a single time per repeat so i'll file that upstream

transport/transport.test.ts
```diff
@@ -764,15 +823,41 @@ describe.each(testMatrix())(
...
+        await onFinished.pop()?.();
+      }
+      await cleanup();
+    });
+    // vitest runs all the `onTestFinished` callbacks after all the repetitions of the test are
+    // done. That is definitely a choice that was made. Instead, we hand-roll it ourselves to avoid
+    // the callbacks from being run concurrently with each other, which causes a ton of mayhem.
+    const onTestFinished = (f: () => Promise<void>) => {
+      onFinished.push(f);
+    };
```
jackyzha0, line 860: is there a way to avoid this `onFinished` boilerplate? i fear we may add a new suite some day and forget to do this

__tests__/e2e.test.ts
```diff
@@ -34,15 +34,46 @@ import { Procedure, ServiceSchema } from '../router';
...
+    });
+    afterEach(async () => {
+      while (onFinished.length > 0) {
+        await onFinished.pop()?.();
+      }
+      await cleanup();
+    });
+    // vitest runs all the `onTestFinished` callbacks after all the repetitions of the test are
+    // done. That is definitely a choice that was made. Instead, we hand-roll it ourselves to avoid
+    // the callbacks from being run concurrently with each other, which causes a ton of mayhem.
```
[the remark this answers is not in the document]
jackyzha0: https://github.com/vitest-dev/vitest/issues/5869

__tests__/disconnects.test.ts
```diff
@@ -20,15 +20,48 @@ import {
...
+        await onFinished.pop()?.();
+      }
+      await cleanup();
+    });
+    // vitest runs all the `onTestFinished` callbacks after all the repetitions of the test are
+    // done. That is definitely a choice that was made. Instead, we hand-roll it ourselves to avoid
+    // the callbacks from being run concurrently with each other, which causes a ton of mayhem.
+    const onTestFinished = (f: () => Promise<void>) => {
+      onFinished.push(f);
+    };
```
masad-frost, line 64: maybe more work and more boilerplate, I wonder if it's better if we can have `tranport.teardown()` in `afterEach` and move `setup` to be called within each test directly?

lhchavez: what would the advantage of that be? in general, the more boilerplate there is, the more likely it is that we're going to hit an issue somewhere.

__tests__/cleanup.test.ts
```diff
@@ -30,6 +30,19 @@ describe.each(testMatrix())(
...
+      while (cleanups.length > 0) {
+        await cleanups.pop()?.();
+      }
+    });
+    // vitest runs all the `onTestFinished` callbacks after all the repetitions of the test are
+    // done. That is definitely a choice that was made. Instead, we hand-roll it ourselves to avoid
+    // the callbacks from being run concurrently with each other, which causes a ton of mayhem.
+    const addCleanup = (f: () => Promise<void> | void) => {
+      cleanups.push(f);
+    };
```
jackyzha0, line 45: can we keep using `onTestFinished` for now while we have `fileParallelism` set to false? if we use the `testFinishesCleanly` on the test scope we shouldn't have any problems as the cleanup order within a suite is synchronous and fine (vitest does properly wait for onTestFinished of each test to finish running before starting the next one)

```ts
import { describe, expect, test } from "vitest";

describe("suite with repeats", () => {
  test("should run a test", ({ onTestFinished }) => {
    console.log("start 1");
    expect(1).toBe(1);
    onTestFinished(async () => {
      await new Promise<void>((resolve) => {
        setTimeout(() => {
          resolve();
        }, 300);
      });
      console.log("onTestFinished 1");
    });
  });

  test("should run a test 2", ({ onTestFinished }) => {
    console.log("start 2");
    expect(1).toBe(1);
    onTestFinished(() => {
      console.log("onTestFinished 2");
    });
  });
});
```

```
stdout | index.test.ts > suite with repeats > should run a test
start 1

stdout | index.test.ts > suite with repeats > should run a test
onTestFinished 1

stdout | index.test.ts > suite with repeats > should run a test 2
start 2
onTestFinished 2
```

https://replit.com/join/uueakwjekd-jzhao-replit

jackyzha0 (dismissed):

__tests__/disconnects.test.ts
masad-frost: I'm just thinking it's more local everywhere

lhchavez: gotcha. i think it's an acceptable tradeoff to have a few fixtures. these truly qualify as fixtures.

jackyzha0 (approved):

PR #178, opened 2024-06-09
[river] Prevent drift between sessions and connections

lhchavez:
## Why

We have a few invariant violations caused because of the fact that during the initial connection establishment, we do not associate the connection with the session. That means that anything that tries to interact with the session and expects the connection to be there (e.g. the grace period timeout) will not see the connection there and proceed to close the session but not the connection!

## What changed

This change adds a new field to the session: `handshakingConn`, which is used to hold the associated handshaking connection but without promoting it to a full connection (which would cause some mayhem with the rest of the state machinery). It still allows the transport to request for it to be closed safely so that even if we are extremely unlucky and the grace period elapses mid-handshake in another session, we can be safe about the global state and won't leave the peer with a dangling connection that cannot be used. Finally, it also ensures to clear any existing grace period callbacks when installing a new one, to avoid the case where a partial connection overwrites an existing grace period timeout, and then the next time the client connects, the forgotten grace period fires and causes mayhem.

## Versioning

### DIFF
--- a/transport/session.ts
+++ b/transport/session.ts
@@ -118,7 +118,7 @@ export interface SessionOptions {
 export class Session<ConnType extends Connection> {
   private codec: Codec;
   private options: SessionOptions;
-  telemetry: TelemetryInfo;
+  readonly telemetry: TelemetryInfo;
 
   /**
    * The buffer of messages that have been sent but not yet acknowledged.
@@ -129,13 +129,18 @@ export class Session<ConnType extends Connection> {
    * The active connection associated with this session
    */
   connection?: ConnType;
+  /**
+   * A connection that is currently undergoing handshaking. Used to distinguish between the active
+   * connection, but still be able to close it if needed.
+   */
+  private handshakingConnection?: ConnType;
   readonly from: TransportClientId;
   readonly to: TransportClientId;
 
   /**
    * The unique ID of this session.
    */
-  id: string;
+  readonly id: string;
 
   /**
    * What the other side advertised as their session ID
@@ -324,7 +329,7 @@ export class Session<ConnType extends Connection> {
     this.ack = seq + 1;
   }
 
-  closeStaleConnection(conn?: ConnType) {
+  private closeStaleConnection(conn?: ConnType) {
     if (this.connection === undefined || this.connection === conn) return;
     this.log?.info(
       `closing old inner connection from session to ${this.to}`,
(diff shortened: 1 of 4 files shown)


transport/session.ts
```diff
@@ -339,26 +344,70 @@ export class Session<ConnType extends Connection> {
     this.cancelGrace();
     this.sendBufferedMessages(newConn);
     this.connection = newConn;
+    this.handshakingConnection = undefined;
```
jackyzha0, line 347:
```suggestion
    // we only call replaceWithNewConnection after
    // having successfully completed a handshake so we clear
    // it here
    this.handshakingConnection = undefined;
```

transport/session.ts
```diff
@@ -339,26 +344,70 @@ export class Session<ConnType extends Connection> {
...
+    }
+    if (this.disconnectionGrace === undefined) return;
     clearTimeout(this.disconnectionGrace);
     this.disconnectionGrace = undefined;
   }
 
+  /**
+   * Used to close the handshaking connection, if set.
+   */
+  closeHandshakingConnection(expectedHanshakingConn?: ConnType) {
```
jackyzha0, line 398:
```suggestion
  closeHandshakingConnection(expectedHandshakingConn?: ConnType) {
```

transport/transport.ts
```diff
@@ -277,7 +284,11 @@ export abstract class Transport<ConnType extends Connection> {
         `session for ${to} already exists but has a different session id (expected: ${session.advertisedSessionId}, got: ${sessionId}), creating a new one`,
         session.loggingMetadata,
       );
-      this.deleteSession(session);
+      this.deleteSession({
+        session,
+        closeHandshakingConnection: handshakingConn !== undefined,
```
jackyzha0, line 289: seems to me like we shouldn't need to pass this in to deleteSession and deleteSession should always try to close its associated handshakingConnection when we delete it?

transport/transport.ts
```diff
@@ -861,9 +890,21 @@ export abstract class ClientTransport<
     }
   }
 
-  protected deleteSession(session: Session<ConnType>) {
+  protected deleteSession({
```
jackyzha0, line 893: thank u for updating these to take objects lol

jackyzha0 (approved):

transport/transport.ts
lhchavez: the reason why i added it is that sometimes this is called while we already have a handshaking connection registered, so if needs to know "oh, if i see this particular instance in the object, then i need to _not_ touch it because it's still being used".

i'll add a comment.

jackyzha0: is there a case in which we are deleting a session but want to keep the underlying handshake conn?

i guess if we have stale session -> handshake to initiate new session we probably want to keep the handshake conn

lhchavez: yes! the case where `getOrCreateSession` is called _from within the handshake handler_. otherwise we will close the connection through which the handshake came through!

transport/transport.ts
lhchavez: o7

PR #179, opened 2024-06-10
[river] Prevent drift between sessions and connections

lhchavez:
## Why

We have a few invariant violations caused because of the fact that during the initial connection establishment, we do not associate the connection with the session. That means that anything that tries to interact with the session and expects the connection to be there  (e.g. the grace period timeout) will not see the connection there and proceed to close the session but not the connection!

## What changed

This change adds a new field to the session: `handshakingConn`, which is used to hold the associated handshaking connection but without promoting it to a full connection (which would cause some mayhem with the rest of the state machinery). It still allows the transport to request for it to be closed safely so that even if we are extremely unlucky and the grace period elapses mid-handshake in another session, we can be safe about the global state and won't leave the peer with a dangling connection that cannot be used. Finally, it also ensures to clear any existing grace period callbacks when installing a new one, to avoid the case where a partial connection overwrites an existing grace period timeout, and then the next time the client connects, the forgotten grace period fires and causes mayhem.

## Versioning

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@replit/river",
-  "version": "0.23.8",
+  "version": "0.23.9",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "@replit/river",
-      "version": "0.23.8",
+      "version": "0.23.9",
       "license": "MIT",
       "dependencies": {
         "@msgpack/msgpack": "^3.0.0-beta2",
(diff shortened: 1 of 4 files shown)


lhchavez:
This is exactly the same as https://github.com/replit/river/pull/178, but due to a rookie mistake, was merged to another branch :facepalm: .

lhchavez:
@replit/micromanager merge accidentally merged the OG PR to another branch instead of `main`.

jackyzha0 (approved):

PR #181, opened 2024-06-10
[dev] Don't share servers among tests

lhchavez:
## Why

We currently set up a single server that is shared among multiple test cases. That's undesirable, since we want isolation in each test.

## What changed

This change brings the isolation back.

## Versioning

### DIFF
--- a/__tests__/cleanup.test.ts
+++ b/__tests__/cleanup.test.ts
@@ -1,12 +1,4 @@
-import {
-  afterAll,
-  assert,
-  describe,
-  expect,
-  test,
-  vi,
-  onTestFinished,
-} from 'vitest';
+import { afterAll, assert, describe, expect, test, vi } from 'vitest';
 import { iterNext } from '../util/testHelpers';
 import {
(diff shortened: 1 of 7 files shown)


__tests__/cleanup.test.ts
```diff
@@ -26,12 +36,22 @@ import { testMatrix } from './fixtures/matrix';
...
+      handshakeOptions?: ServerHandshakeOptions,
+    ) => ServerTransport<Connection>;
     const cleanups: Array<() => Promise<void> | void> = [];
 
+    beforeEach(async () => {
+      const opts = { codec: codec.codec };
+      let cleanup: () => Promise<void> | void;
+      ({ getClientTransport, getServerTransport, cleanup } =
+        await transport.setup({ client: opts, server: opts }));
+      addCleanup(cleanup);
```
jackyzha0, line 53: this should probably just be `afterEach`

jackyzha0 (approved):

lhchavez:
no longer needed

PR #183, opened 2024-06-11
[dev] Don't share servers among tests 

lhchavez:
## Why

We currently set up a single server that is shared among multiple test
cases. That's undesirable, since we want isolation in each test.

## What changed

This change brings the isolation back. We're also giving up on `onTestFinished`.

## Versioning


lhchavez:
no longer needed

PR #185, opened 2024-06-11
Remove destroy state from transport, close is now destroy

masad-frost:
## Why

- Everywhere else in our code base, `close` is permanent and destructive
- There is currently no way of re-opening a transport, and I don't see us adding that

part of the stack at #184

## What changed

- get rid of `destroy` fn and `destroyed` state on transport
- send after `close` now throws an error

## Versioning


jackyzha0 (approved):

Monkatraz (approved):
lgtm

PR #186, opened 2024-06-11
Tie server lifetime to transport

masad-frost:
## Why

It's a little odd that the lifetime of a server can be disconnected from the transport's, I don't see us needing to close each separately.

## What changed

- Don't expose `close` on the server, run the server's cleanup after `transport` emits the closed event introduced in #184 

## Versioning

  - Removed `server.close()`

### DIFF
--- a/router/server.ts
+++ b/router/server.ts
@@ -44,7 +44,6 @@ import { ServerHandshakeOptions } from './handshake';
 export interface Server<Services extends AnyServiceSchemaMap> {
   services: InstantiatedServiceSchemaMap<Services>;
   streams: Map<string, ProcStream>;
-  close(): Promise<void>;
 }
 
 interface ProcStream {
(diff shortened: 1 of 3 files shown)


.replit
```diff
@@ -12,6 +12,10 @@ channel = "stable-23_05"
 localPort = 3000
 externalPort = 80
 
+[[ports]]
```
jackyzha0, line 15: booo

router/server.ts
```diff
@@ -102,6 +101,16 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
     this.transport.addEventListener('sessionStatus', this.onSessionStatus);
     this.log = transport.log;
+
+    this.transport.addEventListener('transportStatus', async ({ status }) => {
+      if (status !== 'closed') {
+        return;
+      }
+
+      this.transport.removeEventListener('message', this.onMessage);
+      this.transport.removeEventListener('sessionStatus', this.onSessionStatus);
```
jackyzha0, line 111: wouldn't these be removed by close anyways

jackyzha0 (approved):

router/server.ts
masad-frost: they might be triggered during cleanup below, before we remove other listeners, not sure.

PR #187, opened 2024-06-11
Remove all transport listeners on close

masad-frost:
## Why

Better cleanup to avoid leaks

## What changed

Introduced `removeAllListeners()` on `EventDispatcher`, and called it in `transport.close`

## Versioning

### DIFF
--- a/transport/transport.ts
+++ b/transport/transport.ts
@@ -544,6 +544,8 @@ export abstract class Transport<ConnType extends Connection> {
       status: this.status,
     });
 
+    this.eventDispatcher.removeAllListeners();
+
     this.log?.info(`manually closed transport`, { clientId: this.clientId });
   }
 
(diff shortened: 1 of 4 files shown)


transport/transport.ts
```diff
@@ -544,6 +544,8 @@ export abstract class Transport<ConnType extends Connection> {
       status: this.status,
     });
 
+    this.eventDispatcher.removeAllListeners();
```
jackyzha0, line 547: maybe we should provide some kind of feedback if we close and there are still listeners

jackyzha0 (approved):
we should edit https://github.com/replit/river/blob/4c8f702f982af195b4a50d1bffbee7fd9302f785/__tests__/fixtures/cleanup.ts#L60 to remove the extra invariant assertions here

transport/transport.ts
masad-frost: like a log? you think it's a symptom of an issue?

masad-frost: I think if someone has a sessionStatus listener, i don't expect them to have to clean it up tbh

PR #188, opened 2024-06-11
Remove service state destructor

masad-frost:
## Why

Feature is got in the way briefly, and decided might as well remove it, I realized that we don't really have a use for this. I don't think people will be spawning multiple servers within the same process. We can re-add this later as needed.

## What changed

Remove feature from https://github.com/replit/river/pull/112

## Versioning


Monkatraz (approved):
makes sense

masad-frost:
> This is helpful for tests i think though

Hmm, in tests you should have direct access to the state if you want to manipulate it in some fashion.

Monkatraz:
> This is helpful for tests i think though

because services are created and dropped in the same context and so we need this for clean up? wonder if it's fine to just let it memory leak

PR #189, opened 2024-06-11
Don't share servers among tests

jackyzha0:
## Why

We currently set up a single server that is shared among multiple test
cases. That's undesirable, since we want isolation in each test.

Supercedes #181 and #183

## What changed

- add missing error listener to websocket on the client
- custom onTestFinished

## Versioning

### DIFF
--- a/__tests__/fixtures/cleanup.ts
+++ b/__tests__/fixtures/cleanup.ts
@@ -132,13 +132,13 @@ export async function testFinishesCleanly({
   vi.useRealTimers();
 }
 
-export const createPostTestChecks = () => {
+export const createPostTestCleanups = () => {
   const cleanupFns: Array<() => Promise<void>> = [];
   return {
-    onTestFinished: (fn: () => Promise<void>) => {
+    addPostTestCleanup: (fn: () => Promise<void>) => {
       cleanupFns.push(fn);
     },
-    postTestChecks: async () => {
+    postTestCleanup: async () => {
       while (cleanupFns.length > 0) {
         await cleanupFns.pop()?.();
       }
--- a/__tests__/fixtures/cleanup.ts
+++ b/__tests__/fixtures/cleanup.ts
@@ -147,3 +147,17 @@ export async function testFinishesCleanly({
 
   vi.useRealTimers();
 }
+
+export const createPostTestChecks = () => {
+  const cleanupFns: Array<() => Promise<void>> = [];
+  return {
+    onTestFinished: (fn: () => Promise<void>) => {
+      cleanupFns.push(fn);
+    },
+    postTestChecks: async () => {
+      while (cleanupFns.length > 0) {
+        await cleanupFns.pop()?.();
+      }
+    },
+  };
+};
--- a/__tests__/fixtures/cleanup.ts
+++ b/__tests__/fixtures/cleanup.ts
@@ -9,21 +9,13 @@ import {
 import { Server } from '../../router';
 import { AnyServiceSchemaMap } from '../../router/services';
 import { testingSessionOptions } from '../../util/testHelpers';
+import { Value } from '@sinclair/typebox/value';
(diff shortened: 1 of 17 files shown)


transport/transport.test.ts
```diff
@@ -14,18 +14,28 @@ import {
...
+    let getClientTransport: TestSetupHelpers['getClientTransport'];
+    let getServerTransport: TestSetupHelpers['getServerTransport'];
+    beforeEach(async () => {
+      const setup = await transport.setup({ client: opts, server: opts });
+      getClientTransport = setup.getClientTransport;
+      getServerTransport = setup.getServerTransport;
+      return async () => {
+        await postTestChecks();
+        await setup.cleanup();
+      };
```
masad-frost, line 35: wonder if we'll get more vitest surprises with this feature lol

__tests__/fixtures/cleanup.ts
```diff
@@ -131,3 +131,17 @@ export async function testFinishesCleanly({
 
   vi.useRealTimers();
 }
+
+export const createPostTestChecks = () => {
+  const cleanupFns: Array<() => Promise<void>> = [];
+  return {
+    onTestFinished: (fn: () => Promise<void>) => {
```
lhchavez, line 138: might be better to not use this name in case there is cargoculting and we accidentally assume certain semantics.

transport/impls/ws/client.ts
```diff
@@ -59,6 +59,10 @@ export class WebSocketClientTransport extends ClientTransport<WebSocketConnectio
       ws.onclose = (evt) => {
         reject(new Error(evt.reason));
       };
+
+      ws.onerror = (err) => {
+        reject(new Error(err.message));
+      };
```
lhchavez, line 65: nice!

lhchavez (approved):
hooray! abstractions!

transport/transport.test.ts
```diff
@@ -14,18 +14,28 @@ import {
...
-    }) => {
+    const { onTestFinished, postTestChecks } = createPostTestChecks();
+    let getClientTransport: TestSetupHelpers['getClientTransport'];
+    let getServerTransport: TestSetupHelpers['getServerTransport'];
+    beforeEach(async () => {
+      const setup = await transport.setup({ client: opts, server: opts });
+      getClientTransport = setup.getClientTransport;
+      getServerTransport = setup.getServerTransport;
+      return async () => {
+        await postTestChecks();
```
masad-frost, line 33: We run some `expects` in here, are those reported when they fail now that we moved them here?

__tests__/fixtures/cleanup.ts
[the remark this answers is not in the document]
masad-frost: i would call this `addPostTestCheck` or something.

transport/transport.test.ts
masad-frost: I think we should move our post-test-checks to be at the end of every test instead of adding them here, like what if the test fails, we don't want to run those checks. I don't see a reason for us to be doing it this way.

masad-frost (changes requested):

masad-frost:
Forgot to leave a specific comment for request:

I don't think we should be running `expect` after the test finishes, we should just inline those in the test, after the test is finished (fails/succeed) we only do cleanups

transport/transport.test.ts
jackyzha0: i did it this way because if the test fails before it hits the end of the test, the cleanup will never run :/

vitest allows assertions in afterEach (beforeEach cleanup is the same as afterEach) and this safely modifies the result of the test

masad-frost: We can decouple those 2. Clean up is one thing, but it doesn't feel like the right place for assertions, assertions should happen in tests. If we want to test our cleanup logic, we should have tests for the cleanup.

masad-frost (approved):

PR #190, opened 2024-06-12
[protocolv2] Refactor server proc streams to look like client implementation

masad-frost:
## Why

I was running into issues with server cleanup when I was implementing abort on the server due to all the different places that can close the stream and such. The implementation was straight forward on the client, so I decided to mimic the client's implementation of "procStream" which is easy to read and extend, as it has less indirection and is self-contained.

## What changed

- `createProcStream` now registers its own listeners for `sessionStatus` and `message` and handles events coming from there internally.
- Listen to `message` events within the server's constructor only to call `createProcStream`

A lot of the code between the frontend and backend stream handling is completely duplicated, we can probably extract it out and re-use it. I'll revisit this later.

### DIFF
--- a/__tests__/cleanup.test.ts
+++ b/__tests__/cleanup.test.ts
@@ -177,10 +177,10 @@ describe.each(testMatrix())(
 
       // ensure we only have one stream despite pushing multiple messages.
       inputWriter.close();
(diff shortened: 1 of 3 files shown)


jackyzha0 (approved):
same comment as the other pr on stream tracking but lgtm otherwise

PR #192, opened 2024-06-13
Check if the span is recording before passing along traceId

ksh-chy:
## Why

By default we will have a non recording span (with the traceid and spanid being zeros) - this pollutes logs. Let's only include this if the span is recording (which will only be the case if they fall into the correct flag) 

## Versioning


jackyzha0 (approved):
u are the goat (but pls run `npm run fix`)

PR #193, opened 2024-06-13
[protocolv2] Implement client abort

masad-frost:
## Why

Client-side implementation for #175 

## What changed

Client calls now accept an `options` parameter (in addition to `init`), the only option is `signal` which is an abort signal so that the caller can abort the stream. Internally nothing crazy, mostly adding code.

### DIFF
--- a/router/client.ts
+++ b/router/client.ts
@@ -16,10 +16,17 @@ import {
   isStreamClose,
   ControlMessageCloseSchema,
   isStreamCloseRequest,
+  isStreamAbort,
 } from '../transport/message';
 import { Static } from '@sinclair/typebox';
 import { nanoid } from 'nanoid';
-import { BaseErrorSchemaType, Err, Result, AnyResultSchema } from './result';
+import {
+  BaseErrorSchemaType,
+  Err,
+  Result,
+  AnyResultSchema,
+  ErrResultSchema,
+} from './result';
 import { EventMap } from '../transport/events';
 import { Connection } from '../transport/session';
 import { Logger } from '../logging';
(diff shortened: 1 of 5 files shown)


masad-frost:
test failures need a rebase because of reusing the transport :D

will get that up later, but the PR is reviewable

router/client.ts
```diff
@@ -301,28 +322,98 @@ function handleProc(
   >(() => {
     transport.sendRequestCloseControl(serverId, streamId);
   });
-  const removeOnCloseListener = outputReader.onClose(() => {
+  outputReader.onClose(() => {
```
masad-frost, line 325: we don't need to remove the listener, since it's cleaned up when the reader is closed anyway, and we're always closed in cleanup!

router/client.ts
```diff
@@ -301,28 +322,98 @@ function handleProc(
...
     span.addEvent('outputReader closed');
-    maybeCleanup();
-  });
 
-  function maybeCleanup() {
-    if (!inputWriter.isClosed() || !outputReader.isClosed()) {
-      return;
```
masad-frost, line 311: we don't need to check these, added the guards at the call site

router/client.ts
```diff
@@ -272,7 +293,7 @@ function handleProc(
     procedureName,
     streamId,
   );
-  let didSessionDisconnect = false;
+  let cleanClose = true;
```
masad-frost, line 296: renamed because it's used in the abort case as well.

transport/transport.ts
```diff
@@ -534,6 +536,17 @@ export abstract class Transport<ConnType extends Connection> {
     });
   }
 
+  sendAbort(to: TransportClientId, streamId: string) {
```
jackyzha0, line 539: mid state-machine refactor and im wondering if we should move this out of transport because this isn't technically a transport level responsibility -- this is just a nice wrapper

jackyzha0 (approved):

transport/transport.ts
masad-frost: sure, we can just move this and all the other control-style messages out, I just thought I'd put it next to the other ones.

jackyzha0: yeah not strongly opinionated, just thought it was weird haha

masad-frost: yeah didn't right

PR #194, opened 2024-06-13
Expose schema diffing function and types

cbrewster:
## Why

We added some tooling for detecting breaking changes between schemas, but we did not export them, so they couldn't be used from outside this package.

## What changed

- Expose `diffServerSchema` and related types

## Versioning

### DIFF
--- a/router/index.ts
+++ b/router/index.ts
@@ -15,6 +15,13 @@ export {
   SerializedServiceSchema,
   SerializedProcedureSchema,
 } from './services';
+export {
+  diffServerSchema,
+  ServerBreakage,
+  ServiceBreakage,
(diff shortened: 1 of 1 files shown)


cbrewster:

* main:
  * **PR #194** 👈

router/index.ts
```diff
@@ -15,6 +15,13 @@ export {
   SerializedServiceSchema,
   SerializedProcedureSchema,
 } from './services';
+export {
+  diffServerSchema,
+  ServerBreakage,
+  ServiceBreakage,
+  ProcedureBreakage,
+  PayloadBreakage,
```
jackyzha0, line 23: these should be exported as type instead

PR #197, opened 2024-06-14
try reporting results in CI

jackyzha0:
## Why

- we should be able to see what test failed at a quick glance

## What changed

- add test reporter to CI

## Versioning


.github/workflows/ci.yaml
```diff
@@ -42,3 +42,9 @@ jobs:
 
       - name: Test
         run: npm test
```
masad-frost, line 44: Btw should we add --bail for CI?

vitest.config.ts
```diff
@@ -10,6 +10,10 @@ export default defineConfig({
     sequence: {
       hooks: 'stack',
     },
+    reporters: process.env.GITHUB_ACTIONS
+      ? ['basic', 'github-actions', 'junit']
```
masad-frost, line 14: Interesting they got their own reporter

masad-frost (approved):

.github/workflows/ci.yaml
jackyzha0: i think having all results is handy and these minutes dont cost us anything? wdyt

masad-frost: Yeah I guess test runs are pretty fast

masad-frost:
I think. You might need a different name for each job or maybe even test run?

PR #198, opened 2024-06-14
Add options to ignore service/procedure removal during diffing

cbrewster:
## Why

Right now we have no way to mark services/procedures as deprecated, so there is no nice path for removing services/procedures.

Until we have a way to properly handle deprecation, we can skip these checks.

## What changed

- Add options to diffing to allow service and procedure removal

## Versioning


cbrewster:

* main:
  * **PR #198** 👈

jackyzha0 (approved):

PR #200, opened 2024-06-14
[protocolv2] Implement server abort

masad-frost:
## Why

Server-side implementation for #175

## What changed

For tests, recommend disabling whitespace diff as i nested tests in more describe blocks

Mostly abort implementation which is symmetrical with the client implementation. The handler context has mechanisms to deal with abort.

- Handler context interface renamed from `ServiceContextWithTransportInfo` is now `ProcedureHandlerContext`.
- Added doc comments to context
- Removed `from` and `to` from context, they can be retrieved from `session`
- Removed `streamId` from context as it's an internal thing
- Added `abortController` to context, this is the controller for which the handler can decide to abort the request
- Added `clientAbortSignal` to context, this is a signal that is triggered when the client sends an abort, or when the client disconnects.

I'm honestly not super happy with the interface, i think it can be confusing to have a controller and a signal that are separate, ideas welcome. One idea I had is to pass an `abort` function to the context, and an abort signal that is triggered when any of happens:
 - Abort by the client.
 - Abort by the procedure handler.
 - Request completed / normal closure.
 - Client disconnected.

with the abort reason indicating what happened. Updating the interface should be easy, but this PR gets us to a place where we can iterate on the interface.

### DIFF
--- a/__tests__/abort.test.ts
+++ b/__tests__/abort.test.ts
@@ -1,6 +1,11 @@
 import { Type } from '@sinclair/typebox';
 import { beforeEach, describe, expect, test, vi } from 'vitest';
-import { Procedure, ServiceSchema, createClient } from '../router';
+import {
+  Procedure,
+  ServiceSchema,
+  createClient,
+  createServer,
+} from '../router';
 import { testMatrix } from './fixtures/matrix';
 import {
   cleanupTransports,
@@ -8,12 +13,14 @@ import {
   waitFor,
 } from './fixtures/cleanup';
 import { EventMap } from '../transport';
-import { ABORT_CODE } from '../router/procedures';
+import { ABORT_CODE, StreamProcedure } from '../router/procedures';
 import { ControlFlags } from '../transport/message';
 import { TestSetupHelpers } from './fixtures/transports';
+import { nanoid } from 'nanoid';
+import { ProcedureHandlerContext } from '../router/context';
 
 describe.each(testMatrix())(
-  'client initiated abort, client tests ($transport.name transport, $codec.name codec)',
+  'client initiated abort ($transport.name transport, $codec.name codec)',
   async ({ transport, codec }) => {
     const opts = { codec: codec.codec };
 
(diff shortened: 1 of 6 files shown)


router/context.ts
```diff
@@ -44,15 +41,52 @@ export interface ServiceContext {}
...
+   * Metadata parsed on the server. See {@link ParsedMetadata}
+   */
   metadata: ParsedMetadata;
+  /**
+   * The session for this stream. Can be used to identify the request
+   * sender via `session.from`, however, we recommend using the
+   * handshake metadata with some auth mechanism to identify the
+   * the request sender.
+   */
+  session: Session<Connection>; // TODO: only expose a subset interface of session
```
Monkatraz, line 62: glorious comments, also I think rather than expose the entire session (or a subset of a session) we could just lift out whatever service's may need from the session directly into the context, like the `from` id. I doubt you need much

router/server.ts
```diff
@@ -387,6 +448,8 @@ class RiverServer<Services extends AnyServiceSchemaMap>
           initMessage,
           async (span): InputHandlerReturn => {
             try {
+              // TODO handle never resolving after cleanup/full close
+              // which would lead to us holding on to the closure forever
```
Monkatraz, line 452: Can you even avoid the memory leak if the handler never resolves? You could have a timeout condition (where the timeout only starts after everything is closed) but the handler would still have a reference to a lot of things in the context which may force it to stick around. Still would be better than stalling forever though

router/context.ts
```diff
@@ -44,15 +41,52 @@ export interface ServiceContext {}
...
+   * handler and notify the client that the stream was aborted.
+   *
+   * Important to note that this controller is owned by the handler, if you
+   * want to listen to aborts coming from the client, you should use the
+   * {@link clientAbortSignal}.
+   *
+   * Aborts are not the same as closing streams gracefully, please refer to
+   * the river documentation to understand the difference between the two concepts.
+   */
+  abortController: AbortController;
```
Monkatraz, line 74: I wonder if making this a function named something like `abortProcedureUngracefully` would encourage the right practices here (not really a throw, though)

Monkatraz (approved):
lgtm, probably wanna wait for Jacky.

and yeah I think there probably is something more intuitive for the abort props, maybe making the two props feel "symmetrical" would help, like with `reader` and `writer`

__tests__/abort.test.ts
```diff
@@ -330,192 +500,356 @@ describe.each(testMatrix())(
...
+              code: ABORT_CODE,
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+              message: expect.any(String),
+            },
+          },
+        ]);
+        expect(clientOutputReader.isClosed());
+        expect(clientInputWriter.isClosed());
+      });
     });
```
jackyzha0, line 853: we can probably simplify this test code into only the e2e path

router/context.ts
jackyzha0, line 62: we can probably also just use `Omit<Session<Connection>, ...>` here so we do it at the type level, idt it matters if we actually strip stuff

jackyzha0 (approved):

router/context.ts
masad-frost: yeah, actually i ended up just keeping `from` in a later PR :D

router/server.ts
masad-frost: we just need to make sure we avoid capturing things that are not needed in a closure, but i think there's no way about it. I think we should just have a log "procedure X didn't exit after Y seconds of stream end" or something.

__tests__/abort.test.ts
masad-frost: :saluting_face:

router/context.ts
masad-frost: I thought about using a function here, but i wanted to give people a way to track their own aborts, so that they can go `abortController.signal`. I think we'll have to see some people use this stuff and iterate

PR #201, opened 2024-06-14
[protocolv2] Treat uncaught handler errors as abort

masad-frost:
## Why

Uncaught handler errors should abort the stream entirely and lead to a full close, right now they lead to a half close which is not ok.

## What changed

- `transport.sendAbort` now accepts an error result as a 3rd argument to allow us to send different payloads with the abort
- when we see a handler exception, we send an abort
- uncaught errors notify the server's input reader


jackyzha0 (approved):
pogg

PR #202, opened 2024-06-14
[protocolv2] replace returned dispose function with ctx.addCleanUp

masad-frost:
## Why

This isn't really a protocolv2 requirement, more of an interface thing, I considered implementing it on `main` but the way things are setup outside of protocolv2 makes it a little more cumbersome. 

If a procedure throws, it has no way to cleanup, we should allow people to register cleanups regardless. In our usage, I also noticed a common pattern of putting cleanups in an array, and then returning a function that calls the cleanup function.

## What changed

Context now has a function `addCleanup` that allows people to register clean up functions. They're called after the stream ends (for any reason), or immediately if the stream already ended.

### DIFF
--- a/__tests__/cleanup.test.ts
+++ b/__tests__/cleanup.test.ts
@@ -1,11 +1,17 @@
-import { assert, beforeEach, describe, expect, test } from 'vitest';
+import { assert, beforeEach, describe, expect, test, vi } from 'vitest';
 import { getIteratorFromStream, iterNext } from '../util/testHelpers';
 import {
   SubscribableServiceSchema,
   TestServiceSchema,
   UploadableServiceSchema,
 } from './fixtures/services';
-import { createClient, createServer } from '../router';
+import {
+  Procedure,
+  ProcedureHandlerContext,
+  ServiceSchema,
+  createClient,
+  createServer,
+} from '../router';
 import {
   advanceFakeTimersBySessionGrace,
   cleanupTransports,
@@ -15,6 +21,9 @@ import {
 } from './fixtures/cleanup';
 import { testMatrix } from './fixtures/matrix';
 import { TestSetupHelpers } from './fixtures/transports';
+import { ControlFlags } from '../transport/message';
+import { Type } from '@sinclair/typebox';
+import { nanoid } from 'nanoid';
 
 describe.each(testMatrix())(
   'procedures should clean up after themselves ($transport.name transport, $codec.name codec)',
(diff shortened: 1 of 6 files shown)


router/server.ts
```diff
@@ -341,17 +341,21 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
-
-      if (procDispose) {
-        procDispose();
-      }
+      handlerCleanups.forEach((c) => {
+        try {
+          c();
+        } catch {
+          // ignore user cleanup errors
+        }
```
Monkatraz, line 356: Do we want to try reporting these somewhere? You could use something like [`AggregateError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError) for reporting

router/context.ts
```diff
@@ -89,4 +89,10 @@ export type ProcedureHandlerContext<State> = ServiceContext & {
    * the river documentation to understand the difference between the two concepts.
    */
   clientAbortSignal: AbortSignal;
+  /**
+   * Lets you add a function that will run when the stream is cleaned up due to
+   * it ending. If a cleanup is added after the stream has ended, it will
+   * run immediately.
+   */
+  addCleanup: (cleanup: () => void) => void;
```
Monkatraz, line 97: Would we need async cleanup ever? (probably only matters if the _order_ of cleanups matters)

__tests__/cleanup.test.ts
```diff
@@ -366,3 +375,114 @@ describe.each(testMatrix())(
...
+    const serverTransport = getServerTransport();
+    const handler = vi.fn<[ProcedureHandlerContext<object>]>();
+    const serverId = 'SERVER';
+    const serviceName = 'service';
+    const procedureName = procedureType;
+
+    const services = {
+      [serviceName]: ServiceSchema.define({
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-explicit-any
+        [procedureType]: (Procedure[procedureType] as any)({
```
Monkatraz, line 412: despicable, that poor linter is just trying to help you!

router/server.ts
```diff
@@ -437,6 +441,16 @@ class RiverServer<Services extends AnyServiceSchemaMap>
       metadata: sessionMeta,
       abortController: handlerAbortController,
       clientAbortSignal: clientAbortController.signal,
+      addCleanup: (c) => {
+        if (inputReader.isClosed() && outputWriter.isClosed()) {
+          // Everything already closed, call cleanup immediately.
+          c();
```
Monkatraz, line 447: Nitpick: If you have a cleanup that throws, `addCleanup` _may or may not throw_, which feels a little odd. My gut is that it should just never throw in the caller just to keep things consistent (what if you're adding multiple cleanups for some reason?)

Monkatraz (approved):
lgtm, some comments about error handling

router/context.ts
Monkatraz: Also would you ever want to _remove_ a cleanup?

router/procedures.ts
```diff
@@ -167,7 +167,7 @@ export interface SubscriptionProcedure<
     context: ProcedureHandlerContext<State>,
     init: Static<Init>,
     output: WriteStream<Result<Static<Output>, Static<Err>>>,
-  ): Promise<(() => void) | void>;
+  ): Promise<void | undefined>;
```
jackyzha0, line 170: why also take undefined here

router/context.ts
jackyzha0: imo cleanup should never be async, we should keep it as sync until we have a very good reason for it to be async

__tests__/fixtures/services.ts
```diff
@@ -241,14 +241,13 @@ export const SubscribableServiceSchema = ServiceSchema.define(
           returnStream.write(Ok({ result: count }));
         });
 
+        ctx.addCleanup(dispose1);
```
jackyzha0, line 244: no concrete actionable thing but i do wonder about discoverability, seems hard to figure out intuitively how to add a cleanup

router/context.ts
jackyzha0: also nit: `onHandlerFinished` maybe, cleanup is ambiguous as to when it runs

jackyzha0 (approved):
would prefer another name like `onHandlerFinished` or something

router/context.ts
masad-frost: yeah we'll see if that comes up, an easy future change

router/procedures.ts
masad-frost: this is "don't return anything" in the typescript world lol, I can add a comment, but can also turn it back to void, it just helped find places where we returned dispose, might be useful when migrating.

__tests__/fixtures/services.ts
masad-frost: Yeah, I guess RTFM? lol, not sure what to do about it. I think this is strictly better in terms of discoverability though

jackyzha0: yeah lgtm

router/context.ts
masad-frost: I think `onHandlerFinished` makes a lot of sense, and it can double as a way of knowing when the stream ended.

router/server.ts
masad-frost: where do you think we should send these? add an event on the server?

Monkatraz: great question, not sure. an event sounds reasonable, although I wonder if we should have something like an exception sink for River

PR #203, opened 2024-06-15
[protocolv2] Handle invalid requests and internal server errors

masad-frost:
## Why

Clients can send invalid requests for many reasons, most commonly due to backwards incompatible server changes, server should handle those and send back a stream abort with it.

## What changed

- Introduced `INTERNAL_RIVER_ERROR` code since some of the errors are invariant violations on the server
- Split out stream request validation from stream handling
- When we see a bad request, we send `INVALID_REQUEST` code with an abort bit
- Made tracing `createHandlerSpan` accept tracing fields explicitly instead of a transport message

some more changes that I'll note inline

### DIFF
--- a/router/context.ts
+++ b/router/context.ts
@@ -1,4 +1,4 @@
-import { Connection, Session } from '../transport/session';
+import { TransportClientId } from '../transport';
 
 /**
  * ServiceContext exist for the purpose of declaration merging
@@ -54,12 +54,9 @@ export type ProcedureHandlerContext<State> = ServiceContext & {
    */
   metadata: ParsedMetadata;
   /**
-   * The session for this stream. Can be used to identify the request
-   * sender via `session.from`, however, we recommend using the
-   * handshake metadata with some auth mechanism to identify the
-   * the request sender.
+   * The ID of the client that sent this request.
    */
-  session: Session<Connection>; // TODO: only expose a subset interface of session
+  from: TransportClientId;
   /**
    * An AbortController for this stream. This is used to abort the stream from the
    * handler and notify the client that the stream was aborted.
--- a/router/server.ts
+++ b/router/server.ts
@@ -8,6 +8,8 @@ import {
   UNEXPECTED_DISCONNECT_CODE,
   AnyProcedure,
   ABORT_CODE,
+  INVALID_REQUEST_CODE,
+  INTERNAL_RIVER_ERROR_CODE,
 } from './procedures';
 import {
   AnyService,
@@ -23,15 +25,19 @@ import {
   isStreamCloseRequest,
   isStreamAbort,
 } from '../transport/message';
-import { ServiceContext, ProcedureHandlerContext } from './context';
-import { Logger } from '../logging/log';
+import {
+  ServiceContext,
+  ProcedureHandlerContext,
+  ParsedMetadata,
+} from './context';
+import { Logger, MessageMetadata } from '../logging/log';
 import { Value } from '@sinclair/typebox/value';
 import { Err, Result, Ok, ErrResultSchema } from './result';
 import { EventMap } from '../transport/events';
 import { Connection } from '../transport/session';
 import { coerceErrorString } from '../util/stringify';
 import { Span, SpanStatusCode } from '@opentelemetry/api';
-import { createHandlerSpan } from '../tracing';
+import { PropagationContext, createHandlerSpan } from '../tracing';
 import { ServerHandshakeOptions } from './handshake';
 import { ReadStreamImpl, WriteStreamImpl } from './streams';
 
@@ -56,6 +62,20 @@ export interface Server<Services extends AnyServiceSchemaMap> {
 }
 
 type InputHandlerReturn = Promise<(() => void) | void>;
+
+interface NewProcStreamInput {
+  procedure: AnyProcedure;
+  procedureName: string;
+  service: AnyService;
+  serviceName: string;
+  sessionMetadata: ParsedMetadata;
+  loggingMetadata: MessageMetadata;
+  streamId: string;
+  controlFlags: number;
+  tracingCtx?: PropagationContext;
+  initPayload: Static<PayloadType>;
+  from: string;
+}
 
 class RiverServer<Services extends AnyServiceSchemaMap>
   implements Server<Services>
@@ -109,10 +129,17 @@ class RiverServer<Services extends AnyServiceSchemaMap>
       }
 
       if (this.openStreams.has(msg.streamId)) {
+        // has its own message handler
         return;
       }
 
-      this.createNewProcStream(msg);
+      const validated = this.validateNewProcStream(msg);
+
+      if (!validated) {
+        return;
+      }
+
+      this.createNewProcStream(validated);
     };
     this.transport.addEventListener('message', handleMessage);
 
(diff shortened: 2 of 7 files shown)


router/server.ts
```diff
@@ -399,10 +395,6 @@ class RiverServer<Services extends AnyServiceSchemaMap>
 
     const onHandlerError = (err: unknown, span: Span) => {
       const errorMsg = coerceErrorString(err);
-      this.log?.error(
```
masad-frost, line 402: stopped logging here, it's an application-level error, clients can choose to log this

router/server.ts
```diff
@@ -558,11 +551,11 @@ class RiverServer<Services extends AnyServiceSchemaMap>
 
         break;
       default:
-        this.log?.warn(
+        this.log?.error(
```
masad-frost, line 554: promoted to error, this should never happen.

router/server.ts
```diff
@@ -582,6 +575,192 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+      );
+
+      return null;
+    }
+
+    const sessionMetadata =
+      this.transport.sessionHandshakeMetadata.get(session);
+    if (!sessionMetadata) {
+      const errMessage = `session doesn't have handshake metadata`;
+      this.log?.error(errMessage, {
```
masad-frost, line 607: promoted this from warn to error, as it's an internal error, we should never hit this in theory.

router/server.ts
```diff
@@ -582,6 +575,192 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+          message: errMessage,
+        }),
+      );
+
+      return null;
+    }
+
+    if (!isStreamOpen(initMessage.controlFlags)) {
+      const errMessage = `can't create a new procedure stream from a message that doesn't have the stream open bit set`;
+      this.log?.warn(errMessage, {
```
masad-frost, line 626: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -582,6 +575,192 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+          message: errMessage,
+        }),
+      );
+
+      return null;
+    }
+
+    if (!initMessage.serviceName) {
+      const errMessage = `missing service name in stream open message`;
+      this.log?.warn(errMessage, {
```
masad-frost, line 646: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -582,6 +575,192 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+          message: errMessage,
+        }),
+      );
+
+      return null;
+    }
+
+    if (!initMessage.procedureName) {
+      const errMessage = `missing procedure name in stream open message`;
+      this.log?.warn(errMessage, {
```
masad-frost, line 666: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -582,6 +575,192 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+          message: errMessage,
+        }),
+      );
+
+      return null;
+    }
+
+    if (!(initMessage.serviceName in this.services)) {
+      const errMessage = `couldn't find service ${initMessage.serviceName}`;
+      this.log?.warn(errMessage, {
```
masad-frost, line 686: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -582,6 +575,192 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+        }),
+      );
+
+      return null;
+    }
+
+    const service = this.services[initMessage.serviceName];
+    if (!(initMessage.procedureName in service.procedures)) {
+      const errMessage = `couldn't find a matching procedure for ${initMessage.serviceName}.${initMessage.procedureName}`;
+      this.log?.warn(errMessage, {
```
masad-frost, line 707: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -582,6 +575,192 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+      );
+
+      return null;
+    }
+
+    const procedure = service.procedures[initMessage.procedureName];
+
+    if (!Value.Check(procedure.init, initMessage.payload)) {
+      const errMessage = `procedure init failed validation`;
+      this.log?.warn(errMessage, {
```
masad-frost, line 729: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -315,23 +287,47 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+          ...Value.Errors(ControlMessagePayloadSchema, msg.payload),
+        ];
+        let errMessage = 'Expected control payload for procedure with no input';
+        if ('input' in procedure) {
+          errMessage =
+            'Expected either control or input payload, validation failed for both';
+          validationErrors.push(...Value.Errors(procedure.input, msg.payload));
+        }
+
+        this.log?.warn(errMessage, {
```
masad-frost, line 319: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -315,23 +287,47 @@ class RiverServer<Services extends AnyServiceSchemaMap>
       }
 
       if (inputReader.isClosed()) {
-        this.log?.error('Received message after input stream is closed', {
+        this.log?.warn('Received message after input stream is closed', {
```
masad-frost, line 290: demoted to warn, it's a client problem.

router/server.ts
```diff
@@ -293,7 +264,8 @@ class RiverServer<Services extends AnyServiceSchemaMap>
             code: ABORT_CODE,
             message: 'Stream aborted, client sent invalid payload',
           });
-          this.log?.error('Got stream abort without a valid protocol error', {
+          this.log?.warn('Got stream abort without a valid protocol error', {
```
masad-frost, line 267: demoted to warn, it's a client problem.

router/context.ts
```diff
@@ -54,12 +55,9 @@ export type ProcedureHandlerContext<State> = ServiceContext & {
...
   metadata: ParsedMetadata;
   /**
-   * The session for this stream. Can be used to identify the request
-   * sender via `session.from`, however, we recommend using the
-   * handshake metadata with some auth mechanism to identify the
-   * the request sender.
+   * The ID of the client that sent this request.
    */
-  session: Session<Connection>; // TODO: only expose a subset interface of session
+  from: TransportClientId;
```
masad-frost, line 60: this feels like the only thing a service implementer would might need from a session

jackyzha0: lgtm

tracing/index.ts
```diff
@@ -125,22 +120,25 @@ export function createProcTelemetryInfo(
 
 export function createHandlerSpan<Fn extends (span: Span) => unknown>(
```
jackyzha0, line 121: any reason for this change? if we have a message isn't it just more convenient to pass it in

router/server.ts
jackyzha0: is this an invariant violation / should we add that as a tag?

router/server.ts
```diff
@@ -315,23 +287,47 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+          errMessage =
+            'Expected either control or input payload, validation failed for both';
+          validationErrors.push(...Value.Errors(procedure.input, msg.payload));
+        }
+
+        this.log?.warn(errMessage, {
+          ...loggingMetadata,
+          clientId: this.transport.clientId,
+          transportMessage: msg,
+          validationErrors,
```
jackyzha0, line 323: same here: maybe we want to tag this with invariant violation?

router/server.ts
jackyzha0: this should be an invariant violation still and we should tag it as such

jackyzha0: not sure how to effectively disambiguate

1. well behaved clients should never cause us to reach this state
2. even in the case of byzantine clients, we should still never hit this state

jackyzha0 (approved):

router/server.ts
masad-frost: I don't know, maybe we should have an agreement on what's an invariant violation philosophically. For me these are things that are akin to 500 errors in HTTP lingo, bad client requests are 400. If we want, we can make `INVALID_REQUEST` middleware on the client, and report those as invariant violation from the client.

jackyzha0: yeah or at least we should tag invalid-request as such with the tags

masad-frost: yeah, we can add it as a tag, not a bad idea

PR #204, opened 2024-06-15
[protocolv2] Ignore follow up messages to a stream after server aborts it

masad-frost:
## Why

When the server aborts a stream (due to invalid request or explicit abort) the client send more messages before it gets the abort. We should make sure this doesn't lead to errors and unnecessary follow up responses.

## What changed

Keep an LRU of aborted streams, if a new message comes in for a stream in our set, we simply ignore the message.

### DIFF
--- a/.replit
+++ b/.replit
@@ -16,6 +16,26 @@ externalPort = 80
 localPort = 24678
 externalPort = 3000
 
+[[ports]]
+localPort = 36235
+externalPort = 3003
+
+[[ports]]
+localPort = 39589
+externalPort = 3002
+
+[[ports]]
+localPort = 44347
+externalPort = 5173
+
+[[ports]]
+localPort = 45035
+externalPort = 6800
+
+[[ports]]
+localPort = 46789
+externalPort = 5000
+
 [languages.eslint]
 pattern = "**{*.ts,*.js,*.tsx,*.jsx}"
 [languages.eslint.languageServer]
--- a/__tests__/abort.test.ts
+++ b/__tests__/abort.test.ts
@@ -819,6 +819,95 @@ describe.each(testMatrix())(
 
         expect(server.openStreams.size).toEqual(0);
       });
+
+      test('tombstones aborted stream', async () => {
+        const clientTransport = getClientTransport('client');
+        const serverTransport = getServerTransport();
+        addPostTestCleanup(() =>
+          cleanupTransports([clientTransport, serverTransport]),
+        );
+        const serverId = 'SERVER';
+
+        const handler = vi
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          .fn<Parameters<StreamProcedure<any, any, any, any, any>['handler']>>()
+          .mockImplementation(
+            () =>
+              new Promise(() => {
+                // never resolves
+              }),
+          );
+        const services = {
+          service: ServiceSchema.define({
+            stream: Procedure.stream({
+              init: Type.Object({}),
+              input: Type.Object({}),
+              output: Type.Object({}),
+              handler,
+            }),
+          }),
+        };
+
+        createServer(serverTransport, services);
+
+        const serverSendAbortSpy = vi.spyOn(serverTransport, 'sendAbort');
+
+        const serverOnMessage = vi.fn<[EventMap['message']]>();
+        serverTransport.addEventListener('message', serverOnMessage);
+
+        const clientOnMessage = vi.fn<[EventMap['message']]>();
+        clientTransport.addEventListener('message', clientOnMessage);
+
+        const streamId = nanoid();
+        clientTransport.send(serverId, {
+          streamId,
+          serviceName: 'service',
+          procedureName: 'stream',
+          payload: {},
+          controlFlags: ControlFlags.StreamOpenBit,
+        });
+
+        await waitFor(() => {
+          expect(handler).toHaveBeenCalledTimes(1);
+        });
+
+        const [ctx] = handler.mock.calls[0];
+        ctx.abortController.abort();
+        // input for the stream should be ignored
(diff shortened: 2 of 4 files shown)


router/server.ts
```diff
@@ -763,6 +779,32 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+      }
+    }
+    this.items.add(item);
+  }
+
+  has(item: string) {
+    return this.items.has(item);
+  }
+}
+
```
Monkatraz, line 807: Maybe should go into it's own file? This file is already quite Lorge

router/server.ts
```diff
@@ -83,6 +83,12 @@ class RiverServer<Services extends AnyServiceSchemaMap>
   private transport: ServerTransport<Connection>;
   private contextMap: Map<AnyService, ServiceContext & { state: object }>;
   private log?: Logger;
+  /**
+   * We create a tombstones for streams aborted by the server
+   * so that we don't hit errors when the client has inflight
+   * requests it sent before it saw the abort.
+   */
+  private serverAbortedStreams: LRUSet;
```
Monkatraz, line 91: Mainly for discussion: is there a reasonable way to do this that doesn't involve tracking aborted streams like this? I mean it's just a limited size set of strings (and LRU is nice here), so it's relatively harmless, but I wonder if for e.g. multiplayer scenarios (remember classrooms with like 100 people in them?) that you could hit the LRU max size due to the sheer number of aborted streams

(probably just sitting the max size to like 1000 would handle this and not really use that much memory, I don't think the ids are particularly huge)

router/server.ts
```diff
@@ -763,6 +779,32 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+    if (this.items.has(item)) {
+      this.items.delete(item);
+    } else if (this.items.size >= this.maxItems) {
+      const first = this.items.values().next();
+      if (!first.done) {
+        this.items.delete(first.value);
+      }
+    }
+    this.items.add(item);
+  }
```
Monkatraz, line 801: feels elegant, nice

Monkatraz (approved):
lgtm

router/server.ts
masad-frost: ty :)

__tests__/abort.test.ts
```diff
@@ -819,6 +819,95 @@ describe.each(testMatrix())(
...
+
+        const handler = vi
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          .fn<Parameters<StreamProcedure<any, any, any, any, any>['handler']>>()
+          .mockImplementation(
+            () =>
+              new Promise(() => {
+                // never resolves
+              }),
+          );
```
jackyzha0, line 839: nit

```suggestion
        const handler = vi.fn(() => new Promise(() => {}))

File: router/server.ts (hunk 2 above)
Comment (jackyzha0) [reply router/server.ts:91]: i think we could also probably do this based off of the session grace? if we clean up after the session grace elapses we know for certain that there isn't any way we can get a message on that stream

**Review by jackyzha0 — CHANGES_REQUESTED** [review]

File: router/server.ts (hunk 2 above)
Comment (masad-frost) [reply router/server.ts:91]: I couldn't think of another way other than tombstones while still reporting unexpected things. I thought abort would be rare enough that it should be ok to have a small number, and yeah, even if we had like 1 million with our typical 18B nanoid that's like barely 1kb

File: __tests__/abort.test.ts (hunk 4 above)
Comment (masad-frost) [reply __tests__/abort.test.ts:839]: I need the types below so that I can get a typed ctx, plus this gives us an easy place to update types as needed.

File: router/server.ts (hunk 2 above)
Comment (masad-frost) [reply router/server.ts:91]: @jackyzha0 i don't think it's worth the complexity of setting up a separate set for every session or another mapping.

Comment (masad-frost) [reply router/server.ts:91]: starting a timeout is pretty expensive especially if we do that for every abort

Comment (masad-frost) [reply router/server.ts:91]: using session status i guess is fine, but still is it worth the extra map to protect against keeping around extra strings? Maybe some consideration around streamId conflicts?

Comment (masad-frost) [reply router/server.ts:91]: i meant 1mb for 1 million ids

Comment (jackyzha0) [reply router/server.ts:91]: hm yeah i was thinking we could just keep streams around until they are disconnected for good (i.e. we only cleanup streams on session disconnect and aborts still keep streams around), wdyt

Comment (Monkatraz) [reply router/server.ts:91]: yeah all-in-all this way of doing it is pretty elegant, no timers, doesn't grow infinitely, etc etc.

it would take a fairly wacky situation for this to break

Comment (jackyzha0) [reply router/server.ts:91]: i'd prefer a never-breaks solution over a probabilistic one where possible and its not too expensive to, i think keeping streams around until session grace elapses isnt that bad and we aren't adding new timers

File: router/server.ts [hunk 5]
```diff
@@ -778,12 +820,18 @@ export function createServer<Services extends AnyServiceSchemaMap>(
   providedServerOptions?: Partial<{
     handshakeOptions?: ServerHandshakeOptions;
     extendedContext?: Omit<ServiceContext, 'state'>;
+    /**
+     * Maximum number of aborted streams to keep track of to avoid
+     * cascading stream errors.
+     */
+    abortedStreamsMaxTombstones: number;
```
masad-frost, line 827: make this optional

.replit
```diff
@@ -16,6 +16,26 @@ externalPort = 80
...
+externalPort = 5173
+
+[[ports]]
+localPort = 45035
+externalPort = 6800
+
+[[ports]]
+localPort = 46789
+externalPort = 5000
+
```
jackyzha0, line 38: oop

jackyzha0 (approved):

PR #206, opened 2024-06-15
Prevent session id mismatches upon reconnects

lhchavez:
## Why

Currently if a client reconnects and finds that the server was restarted without its knowledge, both sides just paper over this fact and carry on. This causes trouble because now the client has yeeted the previous session and internally replaced it with a new, blank session. That's all good, but the client never told the server that this happened, which is bad, because from now on, every time the client reconnects, they will never agree on what the peer's session id should be, and it will cause no transparent reconnects to be able to happen from then on. This probably also causes some invariant violations, too.

## What changed

This change now ensures that when the client detects a mismatch, it proactively removes the session and reconnects the socket.

Fixes: https://linear.app/replit/issue/WS-3736/better-session-hygiene

## Versioning

### DIFF
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -64,8 +64,7 @@ describe.each(testMatrix())(
 
       // test
       const result = await client.test.add.rpc({ n: 3 });
-      assert(result.ok);
-      expect(result.payload).toStrictEqual({ result: 3 });
+      expect(result).toStrictEqual({ ok: true, payload: { result: 3 } });
       await testFinishesCleanly({
         clientTransports: [clientTransport],
         serverTransport,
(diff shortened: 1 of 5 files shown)


jackyzha0 (approved):
it would be helpful if you could write a test case that demonstrates this causing obviously incorrect behaviour in terms of message delivery

i think this probably is replicable if you yank a server and keep the client intact (test will also help us make sure this is actually fixed :))

as a side note, i'd like to change the session<>connection handling logic to be a bit more robust and turn it into a state machine... i'll try to tackle this monday morning

jackyzha0:
specifically... this test should test exactly that https://github.com/replit/river/blob/7c8d01bf503f8ae1ebacef83e0eda803a55cca75/transport/transport.test.ts#L599

lhchavez:
> it would be helpful if you could write a test case that demonstrates this causing obviously incorrect behaviour in terms of message delivery
> 
> i think this probably is replicable if you yank a server and keep the client intact (test will also help us make sure this is actually fixed :))
> 
> as a side note, i'd like to change the session<>connection handling logic to be a bit more robust and turn it into a state machine... i'll try to tackle this monday morning

test was already there! it's what asserts that the sessions match (an invariant we didn't know we needed and failed before this patch). added a bit more message delivery + session stability assertions after that.

PR #210, opened 2024-06-17
Avoid races when setting the grace period

lhchavez:
## Why

We currently have a race where the `onDisconnect` callback can be invoked arbitrarily late. This means that when replacing a new connection, we can legally get the `onDisconnect` callback invoked after we had already established the new connection, causing a) the grace period to be installed after the reconnection was set and b) the grace period could be invoked with a connection already in place!

## What changed

This change tries _even harder_ to avoid races with the session <-> connection mapping. Now whenever `onDisconnect` is called with the wrong parameters (a mismatched connection), or the grace period elapses with a connection already installed, we just bail out and avoid making more damage.

## Versioning

### DIFF
--- a/transport/transport.ts
+++ b/transport/transport.ts
@@ -293,6 +293,18 @@ export abstract class Transport<ConnType extends Connection> {
    * @param connectedTo The peer we are connected to.
    */
   protected onDisconnect(conn: ConnType, session: Session<ConnType>) {
+    if (session.connection !== undefined && session.connection.id !== conn.id) {
+      // it is completely legal for us to receive the onDisconnect notification later down the line
+      // and accidentally try to install the grace notification into an already-reconnected session.
+      session.telemetry.span.addEvent('onDisconnect race');
+      this.log?.warn('onDisconnect race', {
+        clientId: this.clientId,
+        ...session.loggingMetadata,
+        ...conn.loggingMetadata,
+        tags: ['invariant-violation'],
+      });
+      return;
+    }
     conn.telemetry?.span.end();
     this.eventDispatcher.dispatchEvent('connectionStatus', {
       status: 'disconnect',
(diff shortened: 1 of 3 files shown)


transport/transport.ts
```diff
@@ -293,6 +293,17 @@ export abstract class Transport<ConnType extends Connection> {
...
    */
   protected onDisconnect(conn: ConnType, session: Session<ConnType>) {
+    if (session.connection !== undefined && session.connection.id !== conn.id) {
+      // it is completely legal for us to receive the onDisconnect notification later down the line
+      // and accidentally try to install the grace notification into an already-reconnected session.
+      session.telemetry.span.addEvent('onDisconnect race');
+      this.log?.warn('onDisconnect race', {
+        clientId: this.clientId,
+        ...session.loggingMetadata,
+        ...conn.loggingMetadata,
```
jackyzha0, line 303:
```suggestion
        ...conn.loggingMetadata,
        tags: ['invariant-violation'],
```

transport/transport.ts
```diff
@@ -301,6 +312,17 @@ export abstract class Transport<ConnType extends Connection> {
...
     session.connection = undefined;
     session.beginGrace(() => {
+      if (session.connection !== undefined) {
+        // if for whatever reason the session has a connection, it means that we accidentally
+        // installed a grace period in a session that already had reconnected. oops.
+        session.telemetry.span.addEvent('session grace period race');
+        this.log?.warn('session grace period race', {
+          clientId: this.clientId,
+          ...session.loggingMetadata,
+          ...conn.loggingMetadata,
```
jackyzha0, line 322:
```suggestion
          ...conn.loggingMetadata,
          tags: ['invariant-violation'],
```

jackyzha0 (approved):

PR #212, opened 2024-06-18
Make the server reject misbehaving connections

lhchavez:
## Why

We currently have a setup where the server just trusts whatever the client requests, and stomps over whatever was already there. That's okay in some cases, but not all! For example, if the client's timers are all wonky (like they are when the tab is hidden), we might get in a situation where the client tries to reconnect after its server-side grace period has been exceeded, and they don't agree on what the state should be. Also, if pid2 restarts for whatever reason, both parties should quickly agree that they're not agreeing upfront.

## What changed

This change now adds a new (optional) field to the handshake, `expectedState`. The client should fill this up and explicitly tell the server whether this is a reconnect attempt (currently this is done by checking if the `advertisedSessionId` is set, but that can be done by explicitly tracking whether the session has seen the server at least once when we move away from this model), and tell the server what its expected next sequence number is (so that the server can also detect drifts in this). If the clients don't fill this information up, the server will behave in a backwards-compatible fashion for safety.

## Versioning

- [ ] Breaking protocol change :massage_man: (backwards-compatible)
- [ ] Breaking ts/js API change :massage_man:

### DIFF
--- a/transport/transport.ts
+++ b/transport/transport.ts
@@ -205,6 +205,67 @@ export abstract class Transport<ConnType extends Connection> {
     return session;
   }
 
+  protected createNewSession({
+    to,
+    conn,
+    sessionId,
+    propagationCtx,
+  }: {
+    to: TransportClientId;
+    conn: ConnType;
+    sessionId: string;
+    propagationCtx?: PropagationContext;
+  }) {
+    let session = this.sessions.get(to);
+    if (session !== undefined) {
+      // this is a new session. if the client already had one, delete it before continuing
+      this.log?.info(
+        `session for ${to} already exists, replacing it with a new session as requested`,
+        session.loggingMetadata,
+      );
+      this.deleteSession({
+        session,
+        closeHandshakingConnection: false,
+      });
+      session = undefined;
+    }
+
+    session = this.createSession(to, conn, propagationCtx);
+    session.advertisedSessionId = sessionId;
+    this.log?.info(`created new session for ${to}`, session.loggingMetadata);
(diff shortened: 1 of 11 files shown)


transport/transport.ts
```diff
@@ -205,6 +205,67 @@ export abstract class Transport<ConnType extends Connection> {
...
+
+    this.log?.info(
+      `reused existing session for ${to}`,
+      session.loggingMetadata,
+    );
+
+    return session;
+  }
+
   protected getOrCreateSession({
```
masad-frost, line 269: Feels weird to have this and the two methods above, unclear how they're different out of the context of this PR, let's add some comments

masad-frost: I guess this should only be used in the deprecated code path?

masad-frost (approved):
Yay for explicitness, much easier to reason about.

Let's update PROTOCOL.md?

transport/transport.ts
lhchavez: it's used in lots of places :upside_down_face:. since we create sessions implicitly when we send the first message, we're still using this in more codepaths.

it would be super fun if we didn't need this at all, but that would require some more changes to always make connection an explicit step.

masad-frost: Might be worth it to simplify and remove the feature, don't think it's particularly useful, in fact https://github.com/replit/repl-it-web/pull/45142

lhchavez: per Slack, we're on track to yeet it!

PR #213, opened 2024-06-18
Prevent yet another bamboozle

lhchavez:
## Why

Currently due to the fact that we reuse client IDs for different sessions, the server can be bamboozled into accidentally closing the wrong session!

## What changed

This change now adds more checks to prevent the server from accidentally deleting the wrong session, or accidentally replacing one session with another without our knowledge.

## Versioning

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@replit/river",
-  "version": "0.23.14",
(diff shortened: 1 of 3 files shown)


jackyzha0 (approved):
i still dont understand why this would happen but should be ok as a stopgap until we make the state machine

PR #214, opened 2024-06-18
Delete the Naive JSON codec

lhchavez:
## Why

We don't use this and it's only making our test harness slower.

## What changed

This change drops the `NaiveJsonCodec`, which is slower and more verbose.

## Versioning

- [X] Breaking ts/js API change


jackyzha0:
personally I find it really helpful when debugging tests to use JSON mode so i can inspect the raw payload more easily

imo ok to keep around

lhchavez:
> personally I find it really helpful when debugging tests to use JSON mode so i can inspect the raw payload more easily
> 
> imo ok to keep around

imo we shouldn't be testing code that is not being shipped, since it slows everyone down (don't pay for what you don't use!).

if it's about ease of inspection and debugging, we should do something different for that.

Monkatraz (approved):

jackyzha0:
i don't think it considerably slows any thing down? CI is maybe ~5-8s faster but its already fast enough that I don't really notice it

we also dont pay for CI as this is an open source repo

lhchavez:
> i don't think it considerably slows any thing down? CI is maybe ~5-8s faster but its already fast enough that I don't really notice it
> 
> we also dont pay for CI as this is an open source repo

i constantly have to fight against the matrix of tests that need to be re-run every time i make one change in one test case. those 8s and ~40 extra lines of output add up :upside_down_face:. how about we make debugging easier? we can probably make some convenience functions to decode webpack into JSON when things fail.

jackyzha0:
closing for now, we can reconsider later :)

PR #219, opened 2024-06-21
Supply context to initializeState

masad-frost:
## Why

We want to use shared instances across services in their state, specifically `ToolChainWatcher` in our codebase, we're having to create a new one for each service when we should be re-using the same one and instantiating it in the extended context.

## What changed

`initializeState` is now passed `ServiceContext`

## Versioning

backwards compatible


Monkatraz (approved):
wonder if the context should be wrapped in `Partial` for state factories

jackyzha0 (approved):

PR #220, opened 2024-06-25
Use a state machine to manage session state transitions

jackyzha0:
## Why

Keeping track of what handlers are registered when and knowing when to cleanup/register new callbacks and fields in a session's lifecycle is really difficult. Let's encode valid states via a state machine and make illegal states unrepresentable.

This PR also addresses two other problems that are heavily tied to the session lifecycle:
- Message order synchronization during handshake: we used to synchronize `nextExpectedSeq` in the handshake which is useful to detect desyncs in server to client direction (that is, if the server was about to send a message that would cause the client to receive an out-of-order message, we should just close the session there instead of continuing). However, we never did this check in the other direction.
- Relying on browser timers: we used to rely on a client-side `setInterval` to send heartbeat messages to the server and the server would kill the client if it didn't send heartbeats frequently enough. Turns out that Chromium _heavily_ throttles intervals when backgrounded (source: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/scheduler/main_thread/page_scheduler_impl.h;l=54?q=kDefaultThrottledWakeUpInterval&ss=chromium%2Fchromium%2Fsrc) so a `setInterval` for, say 3s, could take up to 60s to actually fire. This meant connections were getting killed far too frequently. However, if we make the timers server authoritative, the client can always respond to server heartbeats as data-listener-based wakeups are not throttled :)

## What changed

- Implement a state machine that supports async-transitions via a callback-based system. We use a proxy to ensure that session states we've transitioned from cannot be used again.
  - Each state registers what to do on state exit and on close (exiting from the state machine entirely, e.g. closing the session object) so we have consistent cleanups that don't leak. This has been the source of quite a few bugs.
- Made heartbeats server authoritative.
  - Clients no longer schedule their own heartbeats. Instead, it responds directly to any server-sent heartbeats.
  - Servers are responsible for closing a connection on too many heartbeat misses.
  - This is ok even in the case of phantom disconnects as closing the underlying TCP conn will still propagate in the case of a WebSocket transport (and UDS cannot have phantom disconnects)
- Update the handshake messages
  - No longer sends `reconnect`, this can be inferred a combination of `nextExpectedSeq` and `nextSentSeq`
  - Now relies on `sessionId` to generate the ID on the session on the server side so the IDs of the client session and the server session match
  - Now sends `nextSentSeq`. We do some more comparison here to check for mismatched session state before calling the handshake ok. Specifically, there are two cases where we cannot accept a session:
	  - the client is in the future (`client.nextSentSeq > server.ack`)
	  - server is in the future (`server.seq > client.nextExpectedSeq`)
- Distinguish retriable handshake failures from fatal ones.
  - Only a session state mismatch is retriable (i.e. kill the session and start over), everything else should be 'fatal' and stop retrying.
- Testing infra changes
  - Made constructing a transport not automatically connect by default. We had cases where we registered listeners _after_ connecting and would miss some events.
  - No longer expose connections to the transport (it doesn't manage these! it shouldn't be able to introspect connections). I pulled these out into test helpers `numberOfConnections(transport)` and `closeAllConnections(transport)` which covers our existing use cases.

## Versioning


transport/server.ts
```diff
@@ -320,15 +320,7 @@ export abstract class ServerTransport<
...
     let isTransparentReconnect: boolean;
-    if (!parsed.payload.expectedSessionState) {
-      // TODO: remove once we have upgraded all clients.
-      ({ session, isTransparentReconnect } = this.getOrCreateSession({
-        to: parsed.from,
-        conn,
-        sessionId: parsed.payload.sessionId,
-        propagationCtx: parsed.tracing,
-      }));
-    } else if (parsed.payload.expectedSessionState.reconnect) {
```
masad-frost, line 331: I guess python-client is deployed? https://github.com/replit/river-python/pull/32

jackyzha0: yeah i figured we've marinated long enough

transport/sessionStateMachine/common.ts
```diff
@@ -0,0 +1,259 @@
...
+
+export const ERR_CONSUMED = `session state has been consumed and is no longer valid`;
+
+/**
+ *                   0. SessionNoConnection         ◄──┐
+ *                   │  reconnect / connect attempt    │
+ *                   ▼                                 │
+ *                   1. SessionConnecting      ────────┤ connect failure
+ *                   │  connect success                │
+ *                   ▼                                 │ handshake failure
```
masad-frost, line 69: doesn't handshake failure send us to a state where the session is to be disposed altogether? should we express that explicitly?

transport/sessionStateMachine/common.ts
```diff
@@ -0,0 +1,259 @@
...
+ *                   2. SessionHandshaking     ────────┤ connection drop
+ * 4. PendingSession │  handshake success              │
+ * │  server-side    ▼                                 │ connection drop
+ * └───────────────► 3. SessionConnected   ────────────┘ heartbeat misses
+ */
```
masad-frost, line 74: noice, maybe move it to the transitions.ts?

transport/sessionStateMachine/common.ts
```diff
@@ -0,0 +1,259 @@
...
+  onConnectionErrored: (err: unknown) => void;
+  onConnectionClosed: () => void;
+  onHandshake: (msg: OpaqueTransportMessage) => void;
+}
+
+export interface SessionConnectedListeners {
+  onConnectionErrored: (err: unknown) => void;
+  onConnectionClosed: () => void;
+  onMessage: (msg: OpaqueTransportMessage) => void;
+}
```
masad-frost, line 41: why not colocate next to the session classes?

transport/sessionStateMachine/transitions.ts
```diff
@@ -0,0 +1,146 @@
...
+    session.ack,
+    session.sendBuffer,
+    session.telemetry,
+    session.options,
+  ];
+}
+
+export const SessionStateMachine = {
+  entrypoints: {
+    NoConnection(
```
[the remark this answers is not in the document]
masad-frost: It'd be nice to express the entrypoints more concretely in terms of client/server.

transport/sessionStateMachine/common.ts
```diff
@@ -0,0 +1,259 @@
...
+
+  // whether this state has been consumed
+  // and we've moved on to another state
+  _isConsumed: boolean;
+
+  // called when we're transitioning to another state
+  abstract _onStateExit(): void;
+
+  // called when we exit the state machine entirely
+  abstract _onClose(): void;
```
masad-frost, line 86: maybe should add a comment to these that they're private to the state transitions

masad-frost, line 86: `on` prefix can be a little confusing here, it usually means you can register an event listener. Maybe `handle` prefix would be clearer. I've used them interchangeably in the past and ended up confusing myself

transport/sessionStateMachine/stateMachine.test.ts
```diff
@@ -0,0 +1,1117 @@
...
+
+    const session = SessionStateMachine.entrypoints.PendingIdentification(
+      'from',
+      conn,
+      listeners,
+      testingSessionOptions,
+    );
+
+    return { session, ...listeners };
+  };
```
masad-frost, line 190: nit: was a little confused that some helpers are outside describe, some inside

transport/sessionStateMachine/index.ts
```diff
@@ -0,0 +1,7 @@
+export { SessionState, Session } from './common';
+export { SessionPendingIdentification } from './SessionPendingIdentification';
+export { SessionConnecting } from './SessionConnecting';
+export { SessionNoConnection } from './SessionNoConnection';
+export { SessionHandshaking } from './SessionHandshaking';
+export { SessionConnected } from './SessionConnected';
+export { SessionStateMachine } from './transitions';
```
masad-frost, line 7: Maybe worth exporting some of the things that aren't supposed to be constructed directly as type exports

transport/sessionStateMachine/common.ts
```diff
@@ -0,0 +1,259 @@
...
+export interface SessionHandshakingListeners {
+  onConnectionErrored: (err: unknown) => void;
+  onConnectionClosed: () => void;
+  onHandshake: (msg: OpaqueTransportMessage) => void;
+}
+
+export interface SessionConnectedListeners {
+  onConnectionErrored: (err: unknown) => void;
+  onConnectionClosed: () => void;
+  onMessage: (msg: OpaqueTransportMessage) => void;
```
masad-frost, line 40: same comment as below re: `on` prefix vs `handle` prefix

`onSomething(handleSomething)`

transport/sessionStateMachine/stateMachine.test.ts
```diff
@@ -0,0 +1,1117 @@
...
+      const listeners = {
+        onConnectionEstablished: vi.fn(),
+        onConnectionFailed: vi.fn(),
+      };
+      session = SessionStateMachine.transition.NoConnectionToConnecting(
+        session,
+        pendingConn,
+        listeners,
+      );
+
```
masad-frost, line 236: maybe worth checking that the listeners are removed?

transport/sessionStateMachine/SessionConnecting.ts
```diff
@@ -0,0 +1,47 @@
+import { Connection } from '../session';
+import {
+  IdentifiedSession,
+  SessionConnectingListeners,
+  SessionState,
+  bestEffortClose,
+} from './common';
+
+export class SessionConnecting<
```
masad-frost, line 9: doc comment pls

transport/sessionStateMachine/SessionConnected.ts
```diff
@@ -0,0 +1,74 @@
+import { PartialTransportMessage } from '../message';
+import { Connection } from '../session';
+import {
+  IdentifiedSession,
+  SessionConnectedListeners,
+  SessionState,
+} from './common';
+
+export class SessionConnected<
```
masad-frost, line 9: doc comment pls

transport/sessionStateMachine/SessionHandshaking.ts
```diff
@@ -0,0 +1,52 @@
+import { TransportMessage } from '../message';
+import { Connection } from '../session';
+import {
+  IdentifiedSession,
+  SessionHandshakingListeners,
+  SessionState,
+} from './common';
+
+export class SessionHandshaking<
```
masad-frost, line 9: doc comment pls

transport/sessionStateMachine/SessionPendingIdentification.ts
```diff
@@ -0,0 +1,58 @@
+import { MessageMetadata } from '../../logging';
+import { TransportMessage } from '../message';
+import { Connection } from '../session';
+import {
+  CommonSession,
+  SessionHandshakingListeners,
+  SessionState,
+} from './common';
+
+export class SessionPendingIdentification<
```
masad-frost, line 10: doc comment pls

masad-frost (approved):
Makes sense at face-value, I would've thought that the transitions are automatically handled within the sessions (e.g. the state machine automatically goes from handshaking to connected when the handshaking session gets a successful handshake), but I guess the transport layer eeds to do some verification in some cases and act as a controller. I think this is fine, but that was my intuition.

transport/sessionStateMachine/common.ts
jackyzha0: oh true lemme edit the diagram

i think its ok to just have the transport be responsible for disposing of the session entirely

transport/sessionStateMachine/transitions.ts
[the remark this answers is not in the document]
jackyzha0: i'll add a docstring for this but tldr;

1. client always starts at NoConnection (its already an identified session)
2. server always starts with PendingIdentification (because we have a connection but no idea who is on the other side)

transport/sessionStateMachine/common.ts
jackyzha0: i think these ones are ok, they follow the behaviour you would expect from onMessage, onData, etc.

i agree that the _onStateExit and _onClose should change though

transport/sessionStateMachine/transitions.ts
```diff
@@ -0,0 +1,282 @@
...
+ *                           │  handshake success       ┌──────┤ connection drop
+ * 5. PendingIdentification  │  handshake failure  ─────┤      │
+ * │  handshake success      ▼                          │      │ connection drop
+ * ├───────────────────────► 4. SessionConnected        │      │ heartbeat misses
+ * │                         │  invalid message  ───────┼──────┘
+ * │                         ▼                          │
+ * └───────────────────────► x. Destroy Session   ◄─────┘
+ *   handshake failure
+ */
+export const SessionStateMachine = {
```
Monkatraz, line 64: mega nitpick: this holds the model/graph for the state machine and is not a state/machine itself - so it being named `SessionStateMachine` feels slightly misleading. maybe `SessionStateModel` or `SessionStateGraph`? mainly wanting to avoid anyone thinking `new SessionStateMachine()` is conceptually valid. this is a mega nitpick tho

transport/sessionStateMachine/common.ts
```diff
@@ -0,0 +1,275 @@
...
+   * Whether this state has been consumed
+   * and we've moved on to another state
+   */
+  _isConsumed: boolean;
+
+  // called when we're transitioning to another state
+  // note that this is internal and should not be called directly
+  // by consumers, the proxy will call this when the state is consumed
+  // and we're transitioning to another state
+  abstract _handleStateExit(): void;
```
Monkatraz, line 48: nitpick: why not `abstract protected handleStateExit(): void;` so that we don't need the underscore convention?

transport/sessionStateMachine/common.ts
```diff
@@ -0,0 +1,275 @@
...
+    this._handleClose();
+  }
+
+  constructor() {
+    this._isConsumed = false;
+
+    // proxy helps us prevent access to properties after the state has been consumed
+    // e.g. if we hold a reference to a state and try to access it after it's been consumed
+    // we intercept the access and throw an error to help catch bugs
+    return new Proxy(this, {
```
Monkatraz, line 65: is the reasoning behind this is that doing this more manually by checking `_isConsumed` in methods is too fragile? (e.g. forgetting to add the check to a subclass method)

transport/sessionStateMachine/SessionPendingIdentification.ts
```diff
@@ -0,0 +1,72 @@
...
+import { CommonSession, SessionState } from './common';
+
+/*
+ * Server-side session that has a connection but is waiting for the client to identify itself.
+ *
+ * Valid transitions:
+ * - PendingIdentification -> NoConnection (on close)
+ * - PendingIdentification -> Connected (on handshake)
+ */
+export class SessionPendingIdentification<
```
Monkatraz, line 14: If a type of session is specific to only client or the server, should we be more explicit and add `Server` or `Client` as a prefix? as-in `ServerSessionPendingIdentification`

router/server.ts
```diff
@@ -217,31 +222,37 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
...
     const outgoing: ProcStream['outgoing'] = pushable({ objectMode: true });
     const needsClose =
       procedure.type === 'subscription' || procedure.type === 'stream';
     const disposables: Array<() => void> = [];
 
+    const wrappedSend = (payload: PartialTransportMessage) => {
+      if (!this.sessionsBeingCleanedUp.has(sessionId)) {
+        this.transport.send(to, payload);
+      }
+    };
```
Monkatraz, line 240: When the session is being cleaned up this quietly ignores whatever was intended to be sent - should it be more explicit? (e.g. throwing).

transport/connection.ts
```diff
@@ -0,0 +1,103 @@
...
+        spanId: spanContext.spanId,
+      };
+    }
+
+    return metadata;
+  }
+
+  _dataListeners = new Set<(msg: Uint8Array) => void>();
+  _closeListeners = new Set<() => void>();
+  _errorListeners = new Set<(err: Error) => void>();
```
Monkatraz, line 34: shouldn't these be private if these have separate getters?

transport/server.ts
```diff
@@ -66,100 +89,350 @@ export abstract class ServerTransport<
...
+    if (parsedMetadata === false) {
+      return;
+    }
+
+    // 4 connect cases
+    // 1. new session
+    //    we dont have a session and the client is requesting a new one
+    //    we can create the session as normal
+    // 2. cliest is reconnecting to an existing session but we don't have it
+    //    reject this handshake, there's nothing we can do to salvage it
```
Monkatraz, line 248: maybe not for this PR, but should we have a mechanism for negotiating creating a new session instead? or is the client just expected to retry and request a new session explicitly?

transport/sessionStateMachine/common.ts
Monkatraz: I'm guessing because the transition code needs to call this

Monkatraz (commented):
not a very thorough look, mainly nitpicks/bikeshedding and questions

router/server.ts
jackyzha0: yup this just does what we were doing before, faris has some changes that obsolete these in v2 so we'll fix it properly there 🔨

transport/sessionStateMachine/common.ts
jackyzha0: yeah, plus the boilerplate code got super annoying to read

transport/sessionStateMachine/common.ts
jackyzha0: yeah transition is outside the class and needs to call it directly, i think its ok the nodejs internals do something similar to this for streams

transport/sessionStateMachine/SessionPendingIdentification.ts
jackyzha0: hm ClientHandshaking and ServerPendingIdentification maybe?

transport/server.ts
jackyzha0: i wanted to keep the handshake fixed-length (e.g. 2-way handshake for now), the intended way is to have the client retry these (we now distinguish retriable and fatal handshake responses so thats the goal there)

transport/sessionStateMachine/common.ts
Monkatraz: maybe it's excessive, but you could do something like:
```ts
function handleStateExit(session: Session) {
  // @ts-expect-error private property
  session.handleStateExit()
}
```
and then make the method `protected` and properly hide it from consumers

Monkatraz: another interesting trick you can use is this:
```ts
class MyClass {
  static doThing(instance: MyClass) {
    // completely valid, no TS error
    instance.privateMethod()
  }
  
  private privateMethod() {}
}
```
quite useful sometimes. static methods can do whatever they want with private props and methods

__tests__/fixtures/cleanup.ts
```diff
@@ -17,21 +20,20 @@ const waitUntilOptions = {
...
 };
 
+export async function advanceFakeTimersByHeartbeat() {
+  await vi.advanceTimersByTimeAsync(testingSessionOptions.heartbeatIntervalMs);
+}
+
 export async function advanceFakeTimersByDisconnectGrace() {
-  for (let i = 0; i < testingSessionOptions.heartbeatsUntilDead; i++) {
-    // wait for heartbeat interval to elapse
-    await vi.runOnlyPendingTimersAsync();
```
jackyzha0, line 23: turns out `runOnlyPendingTimersAsync` introduced incorrect fake timer behaviour so I did my best to remove it

transport/sessionStateMachine/transitions.ts
```diff
@@ -0,0 +1,282 @@
...
+ *                           │  handshake success       ┌──────┤ connection drop
+ * 5. WaitingForHandshake    │  handshake failure  ─────┤      │
+ * │  handshake success      ▼                          │      │ connection drop
+ * ├───────────────────────► 4. SessionConnected        │      │ heartbeat misses
+ * │                         │  invalid message  ───────┼──────┘
+ * │                         ▼                          │
+ * └───────────────────────► x. Destroy Session   ◄─────┘
+ *   handshake failure
+ */
+export const SessionStateGraph = {
```
lhchavez, line 64: dig. is the main reason this is structured this way instead of having the transition functions in the concrete state types so that we can have the transitions in a single place?

lhchavez (commented):
round 1/2 done (pausing for lunch and continuing afterwards)

router/server.ts
```diff
@@ -64,9 +65,13 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
   contextMap: Map<AnyService, ServiceContextWithState<object>>;
   // map of streamId to ProcStream
   streamMap: Map<string, ProcStream>;
-  // map of client to their open streams by streamId
-  clientStreams: Map<TransportClientId, Set<string>>;
-  disconnectedSessions: Set<TransportClientId>;
+
+  // map of sessionId to streamIds
+  sessionToStreamId: Map<string, Set<string>>;
```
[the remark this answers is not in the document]
jackyzha0: oops i forgor

router/server.ts
```diff
@@ -142,25 +146,27 @@ class RiverServer<Services extends AnyServiceSchemaMap> {
     await this.pushToStream(procStream, message, isInitMessage);
   };
 
-  // cleanup streams on session close
   onSessionStatus = async (evt: EventMap['sessionStatus']) => {
-    if (evt.status !== 'disconnect') return;
+    if (evt.status === 'connect') {
+      this.sessionToStreamId.set(evt.session.id, new Set());
```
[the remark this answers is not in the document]
jackyzha0: great idea

transport/impls/uds/connection.ts
```diff
@@ -50,7 +49,7 @@ export class UdsConnection extends Connection {
   }
 
   close() {
-    this.sock.destroy();
+    this.sock.end();
```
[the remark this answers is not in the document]
jackyzha0: i can change it back to destroy but i mostly did this because we see `
Uint32LengthPrefixFraming` throwing `got incomplete message while flushing` rarely and i think its because we destroy immediately which tries to flush midway

transport/connection.ts
```diff
@@ -0,0 +1,103 @@
...
+        spanId: spanContext.spanId,
+      };
+    }
+
+    return metadata;
+  }
+
+  private _dataListeners = new Set<(msg: Uint8Array) => void>();
+  private _closeListeners = new Set<() => void>();
+  private _errorListeners = new Set<(err: Error) => void>();
```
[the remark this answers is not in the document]
jackyzha0: yeah this needs to work on client too and didnt want to polyfill, ill add a comment

transport/sessionStateMachine/index.ts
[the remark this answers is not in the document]
jackyzha0: i can move it to `PROTOCOL.md` as well

transport/sessionStateMachine/transitions.ts
```diff
@@ -0,0 +1,282 @@
...
+            inheritSharedSession(oldSession)
+          : // old session does not exist, create new state
+            [
+              sessionId,
+              from,
+              to,
+              0,
+              0,
+              [],
+              createSessionTelemetryInfo(sessionId, to, from, propagationCtx),
```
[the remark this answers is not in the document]
jackyzha0: yup :) its also why we made the logger optional because short circuiting allows us to avoid constructing expensive objects if its not bound

transport/sessionStateMachine/transitions.ts
jackyzha0: yeah i think it was a lot easier for me to reason about with all the transitions in one place instead of needing to jump between all the states in different files finding what transitioned to what

transport/sessionStateMachine/transitions.ts
```diff
@@ -0,0 +1,282 @@
...
+  },
+  transition: {
+    // happy path transitions
+    NoConnectionToConnecting<ConnType extends Connection>(
+      oldSession: SessionNoConnection,
+      connPromise: Promise<ConnType>,
+      listeners: SessionConnectingListeners,
+    ): SessionConnecting<ConnType> {
+      const carriedState = inheritSharedSession(oldSession);
+      oldSession._handleStateExit();
```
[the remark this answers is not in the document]
jackyzha0: oh interesting, that's neat yeah I'll experiment with that after lunch

__tests__/cleanup.test.ts
```diff
@@ -341,16 +370,15 @@ describe.each(testMatrix())(
 
       // wait for session to disconnect
       clientTransport.reconnectOnConnectionDrop = false;
-      clientTransport.connections.forEach((conn) => conn.close());
+      closeAllConnections(clientTransport);
       await advanceFakeTimersBySessionGrace();
-
-      expect(clientTransport.sessions.size).toEqual(0);
+      await waitFor(() => expect(numberOfConnections(clientTransport)).toBe(0));
```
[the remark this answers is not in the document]
jackyzha0: i believe vitest + jest both allow toBe on primitives https://vitest.dev/api/expect.html#tobe

(it just allowed me to shorten it into a single line because toEqual pushes it to 3 😭)

router/server.ts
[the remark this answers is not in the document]
jackyzha0: actually we dont use TransportClientId here anymore, we used to track a mapping of client id <> stream but it is more accurate to track session id <> stream

transport/sessionStateMachine/SessionWaitingForHandshake.ts
```diff
@@ -0,0 +1,72 @@
...
+    this.handshakeTimeout = setTimeout(() => {
+      listeners.onHandshakeTimeout();
+    }, this.options.handshakeTimeoutMs);
+
+    this.conn.addDataListener(this.onHandshakeData);
+    this.conn.addErrorListener(listeners.onConnectionErrored);
+    this.conn.addCloseListener(listeners.onConnectionClosed);
+  }
+
+  onHandshakeData = (msg: Uint8Array) => {
```
lhchavez, line 41: how are we guaranteeing that once this is called, the handshake timeout will not fire?

transport/sessionStateMachine/SessionWaitingForHandshake.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  sendHandshake(msg: TransportMessage): boolean {
+    return this.conn.send(this.options.codec.toBuffer(msg));
+  }
+
+  _handleStateExit(): void {
+    this.conn.removeDataListener(this.onHandshakeData);
+    this.conn.removeErrorListener(this.listeners.onConnectionErrored);
+    this.conn.removeCloseListener(this.listeners.onConnectionClosed);
+    clearTimeout(this.handshakeTimeout);
```
lhchavez, line 66: sshould this also check if this is `undefined` and clear the timeout from the object once it's cleared?

transport/sessionStateMachine/SessionHandshaking.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+/*
+ * A session that is handshaking and waiting for the other side to identify itself.
+ *
+ * Valid transitions:
+ * - Handshaking -> NoConnection (on close)
+ * - Handshaking -> Connected (on handshake)
+ */
+export class SessionHandshaking<
```
lhchavez, line 22: if this state is only valid for clients (since the corresponding server-side change is `SessionWaitingForHandshake`) should we add a `Client` to this name and a `Server` to the other state?

transport/sessionStateMachine/SessionConnecting.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.connPromise = connPromise;
+    this.listeners = listeners;
+
+    this.connectionTimeout = setTimeout(() => {
+      listeners.onConnectionTimeout();
+    }, this.options.connectionTimeoutMs);
+
+    connPromise.then(
+      (conn) => {
+        if (this._isConsumed) return;
```
lhchavez, line 44: this is the line of code that is the most load-bearing of the whole PR, haha.

transport/sessionStateMachine/SessionConnected.ts
```diff
@@ -0,0 +1,185 @@
...
+    }
+  };
+
+  _handleStateExit(): void {
+    super._handleStateExit();
+    this.conn.removeDataListener(this.onMessageData);
+    this.conn.removeCloseListener(this.listeners.onConnectionClosed);
+    this.conn.removeErrorListener(this.listeners.onConnectionErrored);
+    clearInterval(this.heartbeatHandle);
+    this.heartbeatHandle = undefined;
```
lhchavez, line 178:
```suggestion
    if (this.heartbeatHandle) {
      clearInterval(this.heartbeatHandle);
      this.heartbeatHandle = undefined;
    }
```

idempotency ftw

transport/sessionStateMachine/SessionConnected.ts
```diff
@@ -0,0 +1,185 @@
...
+
+    // discard acks (unless we aren't heartbeating in which case just respond)
+    this.log?.debug(`discarding msg (ack bit set)`, {
+      ...this.loggingMetadata,
+      transportMessage: parsedMsg,
+    });
+
+    // if we are not actively heartbeating, we are in passive
+    // heartbeat mode and should send a response to the ack
+    if (!this.isActivelyHeartbeating) {
```
lhchavez, line 167: is there a way to create a new state for this? it feels like a new state.

transport/client.ts
```diff
@@ -55,428 +58,324 @@ export abstract class ClientTransport<
...
-    if (!reconnectPromise) {
-      // check budget
-      if (!this.retryBudget.hasBudget(to)) {
-        const budgetConsumed = this.retryBudget.getBudgetConsumed(to);
-        const errMsg = `tried to connect to ${to} but retry budget exceeded (more than ${budgetConsumed} attempts in the last ${this.retryBudget.totalBudgetRestoreTime}ms)`;
-        this.log?.error(errMsg, { clientId: this.clientId, connectedTo: to });
-        this.protocolError(ProtocolError.RetriesExceeded, errMsg);
-        return;
-      }
+    // check budget
```
lhchavez, line 275: is it worth having the "sleeping before reconnecting" state be distinct from "no connection"?

lhchavez (approved):
go forth! please prioritize any comments about correctness / invariants. everything else can be done later.

router/server.ts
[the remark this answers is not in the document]
lhchavez: can we make a type for that? so that we don't have to second-guess ourselves about what this ID is for, lol.

__tests__/cleanup.test.ts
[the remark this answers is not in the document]
lhchavez: rofl. i guess we should have consistency even if that causes prettier to do stuff that ain't _as_ pretty.

transport/impls/uds/connection.ts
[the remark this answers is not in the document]
lhchavez: ah! yes, that will definitely drop some stuff on the floor. adding the `'end'` listener is probably safe in the short term.

transport/transport.ts
```diff
@@ -572,18 +164,12 @@ export abstract class Transport<ConnType extends Connection> {
       throw new Error(err);
     }
 
-    return this.getOrCreateSession({ to }).session.send(msg);
-  }
+    let session = this.sessions.get(to);
+    if (!session) {
+      session = this.createUnconnectedSession(to);
```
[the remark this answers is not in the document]
jackyzha0: i'll add a comment here but tldr is that creating a session won't try to connect it immediately, that transition is controlled by `connect` which is a separate call

this allows us to 'buffer' in an unconnected state until connect actually happens

transport/transport.ts
```diff
@@ -614,4 +200,106 @@ export abstract class Transport<ConnType extends Connection> {
   getStatus(): TransportStatus {
     return this.status;
   }
+
+  protected updateSession<S extends Session<ConnType>>(session: S): S {
+    const activeSession = this.sessions.get(session.to);
```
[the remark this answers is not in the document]
jackyzha0: not yet, i wanted to push that to a future PR

transport/sessionStateMachine/SessionConnected.ts
jackyzha0: isn't `clearInterval` already idempotent?

transport/sessionStateMachine/SessionWaitingForHandshake.ts
jackyzha0: great question, right now we are relying on the transport to listen for the onHandshake and then transition out of this state (which clears the handshake timeout)

jackyzha0: i'll add a comment

transport/sessionStateMachine/transitions.ts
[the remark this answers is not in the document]
jackyzha0: we have enough non-standard inheriting behaviour that it doesn't get us that much re: uniformity, leaving as is for now

transport/sessionStateMachine/SessionWaitingForHandshake.ts
jackyzha0: other state transitions don't clear the other listeners as soon as an even is emitted so this is consistent with it (the listener of the event is expected to do the transition which does cleanup)

transport/sessionStateMachine/SessionConnected.ts
jackyzha0: hm we could but i dont think it would add much value, this to me seems like a parameter of a state (like a substate?)

transport/client.ts
jackyzha0: same as above, i thought about this as well but not sure if it gets us anything (behaviorally this is identical to SessionConnecting)

transport/transport.ts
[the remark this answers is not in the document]
jackyzha0: i think clients themselves shouldn't need to care about creating a session before making a procedure call, its ok that this is automatic but agree with comment below about ensuring that the server never calls it, ill split it out

transport/sessionStateMachine/SessionWaitingForHandshake.ts
jackyzha0: same as below, `clearTimeout` is idempotent so i think its fine?

transport/sessionStateMachine/SessionConnected.ts
lhchavez: that's putting a lot of trust that node / random browsers will never, ever for any circumstance accidentally reuse an interval ID.

PR #221, opened 2024-06-25
Jonathanz/grace period deduplication

zhounathan:
## Why

As discussed in [Slack](https://replit.slack.com/archives/C030RJ2PRC7/p1718892937858239), currently sessionDisconnectGraceMs in transport option also doubles as the grace period for waiting for a handshake message. This makes disabling transparent disconnects (for AI chat) have unintended consequences because a handshake wait time of 0 will result in unintended disconnects. 
Current code: [code snippet](https://github.com/replit/river/blob/2df88f12a59d6866c2c2555bfc509326df24ca56/transport/client.ts#L78
)
## What changed
Another **optional** field has been added to transport options: `handshakeGraceMs` that defaults to 500 and now replaces some unintended usage of sessionDisconnectGraceMs in handshake grace timers instead. This change will make setting `sessionDisconnectGraceMs = 0` no longer break connections, effectively disabling transparent reconnects that can optimize unnecessary state tracking in AI chat. 

With this change, setting `handshakeGraceMs=0` will always break connections because client & server waits 0ms for a valid handshake message.

## Versioning

Tests are implemented to test the behavior of setting either sessionDisconnectGrace or handshakeGrace to 0. A default value of 5000 ms is applied to `handshakeGraceMs`. The stability of other tests proved that it being an optional field should not break any compatibility concerns.

### DIFF
--- a/transport/session.ts
+++ b/transport/session.ts
@@ -106,7 +106,7 @@ export interface SessionOptions {
   /**
    * Duration to wait for a handshake to complete before considering the connection dead.
    */
-  handshakeGraceMs?: number;
+  handshakeTimeoutMs?: number;
   /**
    * The codec to use for encoding/decoding messages over the wire
    */
--- a/transport/session.ts
+++ b/transport/session.ts
@@ -104,6 +104,10 @@ export interface SessionOptions {
    */
   sessionDisconnectGraceMs: number;
   /**
+   * Duration to wait for a handshake to complete before considering the connection dead.
+   */
+  handshakeGraceMs?: number;
+  /**
    * The codec to use for encoding/decoding messages over the wire
    */
   codec: Codec;
--- a/transport/session.ts
+++ b/transport/session.ts
@@ -423,6 +423,19 @@ export class Session<ConnType extends Connection> {
     return this.ack;
   }
 
+  /**
+   * Check that the peer's next expected seq number matches something that is in our send buffer
+   * _or_ matches our actual next seq.
+   */
+  nextExpectedSeqInRange(nextExpectedSeq: number) {
+    for (const msg of this.sendBuffer) {
+      if (nextExpectedSeq === msg.seq) {
+        return true;
+      }
+    }
+    return nextExpectedSeq === this.seq;
+  }
+
   // This is only used in tests to make the session misbehave.
   /* @internal */ advanceAckForTesting(by: number) {
     this.ack += by;
--- a/transport/transport.test.ts
+++ b/transport/transport.test.ts
@@ -612,9 +612,9 @@ describe.each(testMatrix())(
       codec: codec.codec,
       heartbeatIntervalMs: 1_000,
       heartbeatsUntilDead: 2,
-      sessionDisconnectGraceMs: 5_000,
-      // set the handshake grace period to 0
-      handshakeGraceMs: 0,
+      sessionDisconnectGraceMs: 10_000,
+      // setting session grace to be higher so that only handshake grace passes
+      handshakeGraceMs: 500,
     };
     let testHelpers: TestSetupHelpers;
     let getClientTransport: TestSetupHelpers['getClientTransport'];
@@ -699,6 +699,9 @@ describe.each(testMatrix())(
       await waitFor(() => expect(clientSessStart).toHaveBeenCalledTimes(1));
       await waitFor(() => expect(clientConnStop).toHaveBeenCalledTimes(1));
       await waitFor(() => expect(clientSessStop).toHaveBeenCalledTimes(0));
+
+      // advance the timer past the handshake grace period
+      await advanceFakeTimersByConnectionBackoff();
 
       // kill old server and make a new transport with the new server
       await testHelpers.restartServer();
--- a/transport/transport.test.ts
+++ b/transport/transport.test.ts
@@ -575,9 +575,7 @@ describe.each(testMatrix())(
 
       const oldClientSession = serverTransport.sessions.get('client');
       const oldServerSession = clientTransport.sessions.get('SERVER');
-      expect(oldClientSession).toMatchObject({
-        id: oldServerSession?.advertisedSessionId,
-      });
+      expect(oldClientSession.id).toBe(oldServerSession?.advertisedSessionId);
       expect(oldServerSession).toMatchObject({
         id: oldClientSession?.advertisedSessionId,
       });
--- a/transport/transport.test.ts
+++ b/transport/transport.test.ts
@@ -664,11 +664,15 @@ describe.each(testMatrix())(
         expect(parse).toHaveBeenCalledTimes(1);
       });
 
+      // timeout the connection that is waiting for an empty promise
       const handshakeGrace = opts.handshakeTimeoutMs ?? 500;
       await vi.advanceTimersByTimeAsync(handshakeGrace + 1);
 
-      // expect no session to have been established due to connection timeout
+      // expect no server session/connection to have been established due to connection timeout
       expect(serverTransport.sessions.size).toBe(0);
+      expect(serverTransport.connections.size).toBe(0);
+      // client should not have successfully established any connections
+      expect(clientTransport.connections.size).toBe(0);
 
       // exhaust the retry budget to result in a protocolError.RetriesExceeded
       await vi.advanceTimersByTimeAsync(handshakeGrace + 1);
--- a/transport/transport.test.ts
+++ b/transport/transport.test.ts
@@ -637,7 +637,7 @@ describe.each(testMatrix())(
       const schema = Type.Unknown();
       const get = vi.fn(async () => ({}));
 
-      const parse = vi.fn(async () => {
+      const parse = vi.fn(() => {
         const promise = new Promise(() => {
           // noop we never want this to return
         });
--- a/transport/transport.test.ts
+++ b/transport/transport.test.ts
@@ -547,7 +547,7 @@ describe.each(testMatrix())(
       heartbeatsUntilDead: 2,
       // set the session disconnect grace to 0 to force a hard reconnect
       sessionDisconnectGraceMs: 0,
-      handshakeGraceMs: 5_000,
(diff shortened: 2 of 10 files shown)


transport/transport.test.ts
```diff
@@ -536,6 +537,205 @@ describe.each(testMatrix())(
...
+      clientTransport.reconnectOnConnectionDrop = false;
+      clientTransport.connections.forEach((conn) => conn.close());
+
+      await waitFor(() => expect(clientConnStart).toHaveBeenCalledTimes(1));
+      await waitFor(() => expect(clientSessStart).toHaveBeenCalledTimes(1));
+      await waitFor(() => expect(clientConnStop).toHaveBeenCalledTimes(1));
+      await waitFor(() => expect(clientSessStop).toHaveBeenCalledTimes(0));
+
+      // advance the timer past the handshake grace period
+      await advanceFakeTimersByConnectionBackoff();
```
jackyzha0, line 704: i think we should just do `await vi.advanceTimersByTime(opts.handshakeGraceMs)` here to make sure we don't drift

transport/transport.test.ts
```diff
@@ -536,6 +537,205 @@ describe.each(testMatrix())(
...
+          'connectionStatus',
+          clientConnHandler,
+        );
+        clientTransport.removeEventListener('sessionStatus', clientSessHandler);
+        await cleanupTransports([clientTransport, serverTransport]);
+      });
+
+      const msg1 = createDummyTransportMessage();
+      const msg1Id = clientTransport.send(serverTransport.clientId, msg1);
+      await expect(
```
jackyzha0, line 675: i'm not sure if this is testing what we want, the handshake is already complete by this point, maybe a better test would look like one of the handshake tests in https://github.com/replit/river/blob/62d2222872f11ff82c86e069b1be1abe8951884f/transport/transport.test.ts#L951

1. we should make a custom handshake that just waits and never resolves
2. try to establish a connection and make sure that connection isn't established
3. wait for the timeout to elapse and make sure the raw connection is eventually closed

transport/session.ts
```diff
@@ -103,6 +103,10 @@ export interface SessionOptions {
    * Duration to wait between connection disconnect and actual session disconnect
    */
   sessionDisconnectGraceMs: number;
+  /**
+   * Duration to wait for a handshake to complete before considering the connection dead.
+   */
+  handshakeGraceMs?: number;
```
masad-frost, line 109: typically these are called `timeout`

```suggestion
  handshakeTimeoutMs?: number;
```

transport/transport.test.ts
```diff
@@ -536,6 +537,153 @@ describe.each(testMatrix())(
   },
 );
 
+describe.each(testMatrix())(
+  'transport disabling transparent reconnect',
```
masad-frost, line 541: If you're doing a test table you should make sure to include some identifier in each iteration. See above `describes`

transport/transport.test.ts
```diff
@@ -536,6 +537,153 @@ describe.each(testMatrix())(
...
+      const parse = vi.fn(async (metadata: unknown) => {
+        // advance the timer past the handshake grace period
+        const handshakeGrace = opts.handshakeGraceMs ?? 5_000;
+        await sleep(handshakeGrace);
+        return {
+          // @ts-expect-error - we haven't extended the global type here
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+          kept: metadata.kept,
+        };
+      });
```
masad-frost, line 653: interesting approach, nice.

I'm surprised the test itself doesn't timeout thought, i believe it times out at 500ms!

transport/transport.test.ts
```diff
@@ -536,6 +537,153 @@ describe.each(testMatrix())(
...
+        await testHelpers.cleanup();
+      };
+    });
+
+    test('handshake grace period of 0 should lead to closed connections', async () => {
+      const schema = Type.Object({
+        kept: Type.String(),
+        discarded: Type.String(),
+      });
+      const get = vi.fn(async () => ({ kept: 'kept', discarded: 'discarded' }));
```
masad-frost, line 638: I understand you copied these from another test, but that other test was actually testing metadata stuff. I'd just remove the data here as it was confusing to me and I thought it was relevant. i.e. 
```suggestion
      const schema = Type.Object({
        kept: Type.String(),
        discarded: Type.String(),
      });
      const get = vi.fn(async () => ({ kept: 'kept', discarded: 'discarded' }));
```
```suggestion
      const schema = Type.Unknown();
      const get = vi.fn(async () => ({}));
```

transport/transport.test.ts
```diff
@@ -536,6 +537,153 @@ describe.each(testMatrix())(
...
+        construct: get,
+      });
+      addPostTestCleanup(async () => {
+        await cleanupTransports([clientTransport, serverTransport]);
+      });
+
+      await waitFor(() => {
+        expect(get).toHaveBeenCalledTimes(1);
+        expect(parse).toHaveBeenCalledTimes(1);
+      });
```
masad-frost, line 670: feels like we should wait for `parse` to exit, and we should wait for the client to get a timeout error (i believe it's the protocolError listener on the transport).

masad-frost (approved):
Looks good to me, the test can be more robust and targeted IMO, lemme know if you need some help.

transport/transport.test.ts
zhounathan: @masad-frost added a lot of changes based on suggestions!

- changed naming to handshakeTimeoutMs
- removed unnecessary schema and confusing logic
- added a check for protocolError.RetriesExceeded
- adjusted retry budget and timeout limit so that it runs faster and within time limit
- added identifiers

transport/transport.test.ts
```diff
@@ -536,6 +538,165 @@ describe.each(testMatrix())(
...
+
+      await waitFor(() => expect(clientTransport.connections.size).toEqual(1));
+      await waitFor(() => expect(serverTransport.connections.size).toEqual(1));
+
+      const oldClientSession = serverTransport.sessions.get('client');
+      const oldServerSession = clientTransport.sessions.get('SERVER');
+      expect(oldClientSession).toMatchObject({
+        id: oldServerSession?.advertisedSessionId,
+      });
```
jackyzha0, line 580:
```suggestion
      expect(oldClientSession.id).toBe(oldServerSession?.advertisedSessionId);
```

seems to read slightly better

transport/transport.test.ts
```diff
@@ -536,6 +538,165 @@ describe.each(testMatrix())(
...
+      getServerTransport = testHelpers.getServerTransport;
+      return async () => {
+        await postTestCleanup();
+        await testHelpers.cleanup();
+      };
+    });
+
+    test('handshake grace period of 0 should lead to closed connections', async () => {
+      const schema = Type.Unknown();
+      const get = vi.fn(async () => ({}));
```
jackyzha0, line 640:
```suggestion
      const get = vi.fn();
```

transport/transport.test.ts
```diff
@@ -536,6 +538,165 @@ describe.each(testMatrix())(
...
+
+      const parse = vi.fn(async () => {
+        const promise = new Promise(() => {
+          // noop we never want this to return
+        });
+        return promise;
+      });
```
jackyzha0, line 647:
```suggestion
      const parse = vi.fn(() => {
        const promise = new Promise(() => {
          // noop we never want this to return
        });
        return promise;
      });
```

don't need async there

transport/transport.test.ts
```diff
@@ -536,6 +538,165 @@ describe.each(testMatrix())(
...
+        schema,
+        construct: get,
+      });
+      addPostTestCleanup(async () => {
+        clientTransport.removeEventListener('protocolError', protocolError);
+        await cleanupTransports([clientTransport, serverTransport]);
+      });
+
+      const protocolError = vi.fn();
+      clientTransport.addEventListener('protocolError', protocolError);
```
jackyzha0, line 663:
```suggestion
			const protocolError = vi.fn();
			clientTransport.addEventListener('protocolError', protocolError);
      addPostTestCleanup(async () => {
        clientTransport.removeEventListener('protocolError', protocolError);
        await cleanupTransports([clientTransport, serverTransport]);
      });
```

ordering for scoping reasons, reads more in order that way

transport/transport.test.ts
```diff
@@ -536,6 +538,165 @@ describe.each(testMatrix())(
...
+      await waitFor(() => {
+        expect(get).toHaveBeenCalledTimes(1);
+        expect(parse).toHaveBeenCalledTimes(1);
+      });
+
+      const handshakeGrace = opts.handshakeTimeoutMs ?? 500;
+      await vi.advanceTimersByTimeAsync(handshakeGrace + 1);
+
+      // expect no session to have been established due to connection timeout
+      expect(serverTransport.sessions.size).toBe(0);
```
jackyzha0, line 674: we should try to get a handle to the underlying connection and make sure thats closed too

transport/transport.test.ts
```diff
@@ -536,6 +538,165 @@ describe.each(testMatrix())(
...
+        expect(protocolError).toHaveBeenCalledWith(
+          expect.objectContaining({
+            type: ProtocolError.RetriesExceeded,
+          }),
+        );
+      });
+
+      // cleanup: bring client side connections down and stop trying to reconnect
+      clientTransport.reconnectOnConnectionDrop = false;
+      clientTransport.connections.forEach((conn) => conn.close());
```
jackyzha0, line 690: this should be done in cleanup already, ok to omit

```suggestion
```

jackyzha0 (approved):

transport/transport.test.ts
zhounathan: reverting this due to a lint warning:
`transport/transport.test.ts:578:14 - error TS18048: 'oldClientSession' is possibly 'undefined'.`

transport/transport.test.ts
zhounathan: added two checks for client and server connections

PR #222, opened 2024-06-26
Remove compatibility checking code

cbrewster:
## Why

We've decided to move these internally into Replit's codebase since we want to tailor the compat checking closer to Replit's needs which may not make sense for everyone.

## What changed

- Remove compat checking code

## Versioning


cbrewster:

* main:
  * **PR #222** 👈

masad-frost (approved):
Rip

PR #230, opened 2024-07-10
[protocolv2] Make server backwards compatible for now

masad-frost:
## Why

It's easier to roll out this way.

## What changed

- Added `protocolVersion` to transport
- Added allow-list for accepted protocol versions
- Use `protocolVersion` in the server router to determine how we should deal with closes and aborts

### DIFF
--- a/.replit
+++ b/.replit
@@ -16,50 +16,6 @@ externalPort = 80
 localPort = 24678
 externalPort = 3000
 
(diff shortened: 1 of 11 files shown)


masad-frost:
testing via river-babel

jackyzha0 (approved):

PR #232, opened 2024-07-11
Allow empty client-streams

lhchavez:
## Why

We currently don't force streams to have an initialization, which means that it is possible for clients to end a stream without having sent any messages. This confuses the server!

## What changed

This change now allows clients to send empty streams while the protocolv2 change lands.

## Versioning

### DIFF
--- a/__tests__/handler.test.ts
+++ b/__tests__/handler.test.ts
@@ -69,6 +69,19 @@ describe('server-side test', () => {
     expect(output.readableLength).toBe(0);
   });
 
+  test('stream empty', async () => {
+    const [input, output] = asClientStream(
+      { count: 0 },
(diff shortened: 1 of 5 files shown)


__tests__/handler.test.ts
```diff
@@ -69,6 +69,19 @@ describe('server-side test', () => {
     expect(output.readableLength).toBe(0);
   });
 
+  test('stream empty', async () => {
+    const [input, output] = asClientStream(
```
jackyzha0, line 73: can you do an e2e one too? the handler ones don't exercise the router path

jackyzha0 (approved):
oops

masad-frost:
cries in v2

PR #233, opened 2024-07-12
simplify proxy, move seq check to no session path

jackyzha0:
## Why

- we hit codepaths where we hit a hard reconnect route and skip over the check where we make sure that the client isn't asking for a session we don't have

## What changed

- simplify ws<>uds proxy to not need to introspect the messages
- fix the `else` check in the server so that we properly check `clientNextSentSeq > 0 || clientNextExpectedSeq > 0` even if we just went through a hard reconnect
- change the transparent reconnect test to actually test the right thing

## Versioning


transport/transport.test.ts
```diff
@@ -746,10 +744,9 @@ describe.each(testMatrix())(
       expect(oldClientSessionId).not.toBeUndefined();
       expect(oldServerSessionId).not.toBeUndefined();
 
-      // make sure our connection is still intact even after session grace elapses
-      await advanceFakeTimersBySessionGrace();
```
lhchavez, line 750: hmm wait, this was supposed to be testing something else: the bug where _sometimes_ if you waited long enough after a session establishment, the old grace period timer elapsed and would yank your connection. can we keep that somehow?

(alternatively doing a state machine transition fuzz test would suffice)

transport/transport.test.ts
```diff
@@ -733,10 +733,8 @@ describe.each(testMatrix())(
         await cleanupTransports([clientTransport, serverTransport]);
       });
 
-      await waitFor(() => expect(onConnect).toHaveBeenCalledTimes(1));
-      closeAllConnections(clientTransport);
       await waitFor(() => {
-        expect(onConnect).toHaveBeenCalledTimes(2);
```
lhchavez, line 739: same as the comment below. it's probably better to create a different test.

transport/transport.test.ts
jackyzha0: we have another test for this! `heartbeats should not interrupt normal operation`

i just trimmed the responsibility of the test to do what it says on the tin

lhchavez (approved):

transport/transport.test.ts
```diff
@@ -172,6 +172,39 @@ describe.each(testMatrix())(
...
+        expect(numberOfConnections(serverTransport)).toBe(1);
+        expect(numberOfConnections(clientTransport)).toBe(1);
+      });
+
+      const oldClientSessionId = serverTransport.sessions.get('client')?.id;
+      const oldServerSessionId = clientTransport.sessions.get('SERVER')?.id;
+      expect(oldClientSessionId).not.toBeUndefined();
+      expect(oldServerSessionId).not.toBeUndefined();
+
+      await advanceFakeTimersBySessionGrace();
```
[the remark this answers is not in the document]
jackyzha0: other test "reconnecting before grace period ends should leave session intact" should cover this

PR #234, opened 2024-07-12
Get rid of the UDS transport

lhchavez:
## Why

The UDS transport had a good run, but we're only using WebSockets (over UDS, lol) now in production.

## What changed

This change gets rid of unused code to avoid making more problems for us.

## Versioning

- [X] Breaking ts/js API change (removes stuff!)


lhchavez:
won't merge this until the udsless flag is 100%

jackyzha0 (approved):
🧼

PR #235, opened 2024-07-13
Make damn sure that we only send a single close

lhchavez:
## Why

We are still seeing places where we have client procedure calls that end up closing the stream twice.

## What changed

This change now makes _really_ sure that streams are only closed once. Also adds some machinery to detect any invariant violations during tests and fail. That way we never end up in a situation where we accidentally proceed with a test if it would have caused worsening of stuff.

## Versioning

### DIFF
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -391,7 +391,7 @@ describe.each(testMatrix())(
       });
     });
 
-    test.only('subscription idempotent close', async () => {
+    test('subscription idempotent close', async () => {
       // setup
       const clientTransport = getClientTransport('client');
       const serverTransport = getServerTransport();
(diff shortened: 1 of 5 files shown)


masad-frost (approved):
this is fixed in v2, if we're targetting pid2 clients with this change it can wait, I'm half-way through migrating pid2, if it's ai-chat should merge it.

__tests__/e2e.test.ts
```diff
@@ -352,6 +391,46 @@ describe.each(testMatrix())(
       });
     });
 
+    test.only('subscription idempotent close', async () => {
```
masad-frost, line 394:
```suggestion
    test('subscription idempotent close', async () => {
```

masad-frost:
might as well merge this, i guess

PR #236, opened 2024-07-13
[protocolv2] Fix passing input as first message being backwards incompatible

masad-frost:
## Why

In v1.1 sometimes the first message is not `init`, but instead it's the `input`. Let's handle that.

## What changed

If we get an `init` that's not matching what we expect, we'll check if the client is on v1.1 and sent us an `input` as a first message, then we'll pass the first message as `input` and treat `init` as an empty object (which is what the first pass of requiring `init` for `stream` and `upload` be).

## Versioning

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@replit/river",
-  "version": "0.200.0-rc.4",
+  "version": "0.200.0-rc.6",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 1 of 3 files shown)


masad-frost:
Will link river-babel test PR shortly

masad-frost:
test here https://github.com/replit/river-babel/pull/38/commits/814fe3be94f9d57850821a9d870315dbe8e00099

PR #237, opened 2024-07-15
add backing off state, add test to make sure backoff doesn't count for grace period

jackyzha0:
## Why

- because the Connecting state has a connection timeout, we hit cases in production where the backoff timer _exceeded_ the connection timeout which meant that after a certain point where backoff > connection timeout, we'd never successfully establish a connection

## What changed

- simplify the budget tracking now that client cannot connect to more than one server
- wake up babe, new state machine state just dropped (`SessionBackingOff` which sits between `SessionNoConnection` and `SessionConnecting`)
- add a test to make sure that backoff is not counted towards session grace period
- change session prop passing to be object based rather than long list of params

## Versioning

### DIFF
--- a/transport/rateLimit.ts
+++ b/transport/rateLimit.ts
@@ -1,5 +1,3 @@
-import { TransportClientId } from './message';
-
 /**
  * Options to control the backoff and retry behavior of the client transport's connection behaviour.
  *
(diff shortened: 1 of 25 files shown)


transport/server.ts
```diff
@@ -358,6 +358,15 @@ export abstract class ServerTransport<
             },
           });
 
+        oldSession = noConnectionSession;
+      } else if (oldSession.state === SessionState.BackingOff) {
```
masad-frost, line 362: why does server have this state? :thinking: it seems like a client concern?

transport/sessionStateMachine/SessionConnected.ts
```diff
@@ -102,7 +107,10 @@ export class SessionConnected<
     }, this.options.heartbeatIntervalMs);
   }
 
-  waitForNextHeartbeat() {
+  // kill the connection if we don't receive a heartbeat in time
+  // realistically, this only happens in a proxy scenario where the connection might
+  // be desynchronized
+  startPassiveHeartbeatCheck() {
```
masad-frost, line 113: btw this function might be a performance drain starting so many timeouts and clearing them on every message. Idk how bad, but just something to keep in mind, we can probably have a system that doesn't create that many objects if we notice it's a problem.

transport/rateLimit.ts
```diff
@@ -43,23 +41,21 @@ export interface ConnectionRetryOptions {
 }
 
 export class LeakyBucketRateLimit {
-  private budgetConsumed: Map<TransportClientId, number>;
-  private intervalHandles: Map<
-    TransportClientId,
-    ReturnType<typeof setInterval>
-  >;
+  private budgetConsumed: number;
+  private intervalHandle?: ReturnType<typeof setInterval>;
```
lhchavez, line 45: yessss! at-most-one!

lhchavez (approved):

PR #238, opened 2024-07-16
Allow the custom handler to return more error codes

lhchavez:
## Why

Previously the only allowed error in a custom handler is `false` (which doesn't say a lot). But we want to be able to return something more expressive.

## What changed

This change now adds a new type for this, and allows the custom handler to return more responses. This allows us to have different behavior for different rejection kinds (some more permanent than others).

## Versioning

- [X] Breaking ts/js API change (return type from the validator changed)

### DIFF
--- a/transport/server.ts
+++ b/transport/server.ts
@@ -3,6 +3,7 @@ import { ParsedMetadata } from '../router/context';
 import { ServerHandshakeOptions } from '../router/handshake';
 import {
   ControlMessageHandshakeRequestSchema,
+  HandshakeErrorCustomHandlerFatalResponseCodes,
   HandshakeErrorResponseCodes,
   OpaqueTransportMessage,
   PROTOCOL_VERSION,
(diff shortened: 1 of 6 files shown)


transport/server.ts
```diff
@@ -453,52 +454,55 @@ export abstract class ServerTransport<
...
-        );
+    const parsedMetadata = await this.handshakeExtensions.validate(
+      rawMetadata,
+      previousParsedMetadata,
+    );
 
-        return false;
-      }
+    // handler rejected the connection
+    if (typeof parsedMetadata === 'string') {
```
jackyzha0, line 490:
```suggestion
    if (Value.Check(HandshakeErrorCustomHandlerFatalResponseCodes, parsedMetadata)) {
```

jackyzha0 (commented):
hm river itself shouldn't need to know about `REJECTED_UNSUPPORTED_CLIENT_VERSION` i dont think, can we do this entirely within our custom handshake?

transport/server.ts
masad-frost, line 490: this is tricky, i think we should expect users to return result here to avoid any ambiguity, `parsedMetadat` is of type `any` which could also be a string

lhchavez: Jacky's suggestion handles it better!

lhchavez: :yiss:

lhchavez:
> hm river itself shouldn't need to know about `REJECTED_UNSUPPORTED_CLIENT_VERSION` i dont think, can we do this entirely within our custom handshake?

made it a bit more open-ended: unsupported client. unfortunately we need to have river know _something_ about this case so that it can cross the server/client boundary. we can do it like this with one more enum value, or we can do something more heavy-handed like having the server return a custom payload and let the client decode this custom payload. that's an extensibility knob that feels too much for this.

jackyzha0 (approved):
yeah i think that makes sense, epic 👍

transport/server.ts
masad-frost: I don't think so. The metadata can be string :)

PR #239, opened 2024-07-16
Also ship the handshake failure code

lhchavez:
## Why

We have the handshake failure _reason_, but it would be nice to also have the code to avoid having to do string matching on the reason.

## What changed

This change now adds the `code` when the protocol error is a handshake failure.

## Versioning

### DIFF
--- a/transport/events.ts
+++ b/transport/events.ts
@@ -1,5 +1,6 @@
+import { type Static } from '@sinclair/typebox';
 import { Connection } from './connection';
-import { OpaqueTransportMessage } from './message';
+import { OpaqueTransportMessage, HandshakeErrorResponseCodes } from './message';
 import { Session, SessionState } from './sessionStateMachine';
 import { TransportStatus } from './transport';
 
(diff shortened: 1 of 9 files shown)


transport/events.ts
```diff
@@ -24,10 +25,19 @@ export interface EventMap {
...
+        code: Static<typeof HandshakeErrorResponseCodes>;
+        message: string;
+      }
+    | {
+        type: Omit<
+          ProtocolErrorType,
+          (typeof ProtocolError)['HandshakeFailed']
+        >;
+        message: string;
+      };
```
jackyzha0, line 40:
```suggestion
  protocolError:
    | {
        type: ProtocolErrorType['HandshakeFailed'];
        code: Static<typeof HandshakeErrorResponseCodes>;
        message: string;
      }
    | {
        type: Omit<
          ProtocolErrorType,
          ProtocolErrorType['HandshakeFailed']
        >;
        message: string;
      };
```

does this work?

jackyzha0 (approved):

transport/events.ts
lhchavez: nope, `ProtocolErrorType` is a union of strings, not an enum:
image

PR #240, opened 2024-07-17
add `onSessionGracePeriodElapsed` to `SessionBackingOff`, `SessionConnecting`, and `SessionHandshaking`

jackyzha0:
## Why

- session should track its grace period even as it transitions between various non-connected states
- otherwise, we end up in scenarios where the reconnect loop will prevent the session grace from kicking in until the retry budget is exhausted

## What changed

- add `onSessionGracePeriodElapsed` to intermediate states
- add many many tests for behavior of session grace between states

## Versioning

### DIFF
--- a/router/context.ts
+++ b/router/context.ts
@@ -41,7 +41,7 @@ export interface ServiceContext {}
  * ```
  */
 /* eslint-disable-next-line @typescript-eslint/no-empty-interface */
-export interface ParsedMetadata {}
+export interface ParsedMetadata extends Record<string, unknown> {}
 
 /**
  * The {@link ServiceContext} with state. This is what is passed to procedures.
--- a/transport/server.ts
+++ b/transport/server.ts
@@ -501,20 +501,23 @@ export abstract class ServerTransport<
       ? this.sessionHandshakeMetadata.get(existingSession.to)
       : undefined;
 
-    const parsedMetadata = await this.handshakeExtensions.validate(
+    const parsedMetadataOrFailureCode = await this.handshakeExtensions.validate(
       rawMetadata,
       previousParsedMetadata,
     );
 
     // handler rejected the connection
     if (
-      Value.Check(HandshakeErrorCustomHandlerFatalResponseCodes, parsedMetadata)
+      Value.Check(
+        HandshakeErrorCustomHandlerFatalResponseCodes,
+        parsedMetadataOrFailureCode,
+      )
     ) {
       this.rejectHandshakeRequest(
         handshakingSession,
         from,
         'rejected by handshake handler',
-        parsedMetadata,
+        parsedMetadataOrFailureCode,
         {
           ...handshakingSession.loggingMetadata,
           connectedTo: from,
@@ -525,6 +528,6 @@ export abstract class ServerTransport<
       return false;
     }
 
-    return parsedMetadata;
+    return parsedMetadataOrFailureCode;
   }
 }
--- a/transport/server.ts
+++ b/transport/server.ts
@@ -301,7 +301,11 @@ export abstract class ServerTransport<
       msg.payload.expectedSessionState.nextExpectedSeq;
     const clientNextSentSeq = msg.payload.expectedSessionState.nextSentSeq ?? 0;
 
-    if (oldSession && oldSession.id === msg.payload.sessionId) {
+    if (
+      this.options.enableTransparentSessionReconnects &&
+      oldSession &&
+      oldSession.id === msg.payload.sessionId
+    ) {
       connectCase = 'transparent reconnection';
 
       // invariant: ordering must be correct
@@ -382,10 +386,15 @@ export abstract class ServerTransport<
       // we don't have a session, but the client is trying to reconnect
       // to an old session. we can't do anything about this, so we reject
       connectCase = 'unknown session';
+
+      const rejectionMessage = this.options.enableTransparentSessionReconnects
+        ? `client is trying to reconnect to a session the server don't know about: ${msg.payload.sessionId}`
+        : `client is attempting a transparent reconnect to a session but the server does not support it: ${msg.payload.sessionId}`;
+
       this.rejectHandshakeRequest(
         session,
         msg.from,
-        `client is trying to reconnect to a session the server don't know about: ${msg.payload.sessionId}`,
+        rejectionMessage,
         'SESSION_STATE_MISMATCH',
         {
           ...session.loggingMetadata,
--- a/transport/sessionStateMachine/SessionConnected.ts
+++ b/transport/sessionStateMachine/SessionConnected.ts
@@ -129,7 +129,10 @@ export class SessionConnected<
 
   onMessageData = (msg: Uint8Array) => {
     const parsedMsg = this.parseMsg(msg);
-    if (parsedMsg === null) return;
+    if (parsedMsg === null) {
+      this.listeners.onInvalidMessage('could not parse message');
+      return;
+    }
 
     // check message ordering here
     if (parsedMsg.seq !== this.ack) {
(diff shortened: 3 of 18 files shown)


transport/sessionStateMachine/SessionWaitingForHandshake.ts
```diff
@@ -1,13 +1,22 @@
 import { MessageMetadata } from '../../logging';
 import { Connection } from '../connection';
-import { TransportMessage } from '../message';
-import { SessionHandshakingListeners } from './SessionHandshaking';
+import { OpaqueTransportMessage, TransportMessage } from '../message';
 import { CommonSession, CommonSessionProps, SessionState } from './common';
 
+export interface SessionWaitingForHandshakeListeners {
```
jackyzha0, line 6: disambiguate the listeners from handshake ones for more explicitness (plus this state doesn't need to care about session grace because its not a true identified session yet, we rely on `onHandshakeTimeout` to close this if its open for too long without something happening)

jackyzha0:
todo:
- figure out if we should keep the `set the session disconnect grace to 0 to force a hard reconnect` behaviour or if we should make an explicit session option for this
- add a test for ensuring that session grace is elapsed even in a retry loop

lhchavez (commented):
looks decent!

transport/server.ts
```diff
@@ -382,10 +386,15 @@ export abstract class ServerTransport<
       // we don't have a session, but the client is trying to reconnect
       // to an old session. we can't do anything about this, so we reject
       connectCase = 'unknown session';
+
+      const rejectionMessage = this.options.enableTransparentSessionReconnects
+        ? `client is trying to reconnect to a session the server don't know about: ${msg.payload.sessionId}`
+        : `client is attempting a transparent reconnect to a session but the server does not support it: ${msg.payload.sessionId}`;
```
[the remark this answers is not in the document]
jackyzha0: this is mostly a failsafe on the server side because transparent connects is a client-side setting for the most part

in any case, the client already deletes the session on getting this code in any form so its fine (equivalent to defaulting to hard reconnect behaviour)

transport/sessionStateMachine/SessionConnected.ts
```diff
@@ -129,7 +129,10 @@ export class SessionConnected<
 
   onMessageData = (msg: Uint8Array) => {
     const parsedMsg = this.parseMsg(msg);
-    if (parsedMsg === null) return;
+    if (parsedMsg === null) {
+      this.listeners.onInvalidMessage('could not parse message');
```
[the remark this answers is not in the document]
jackyzha0: `parseMsg` logs this already, i think its ok for the other side not to have this info

```ts
parseMsg(msg: Uint8Array): OpaqueTransportMessage | null {
  const parsedMsg = this.options.codec.fromBuffer(msg);

  if (parsedMsg === null) {
    const decodedBuffer = new TextDecoder().decode(Buffer.from(msg));
    this.log?.error(
      `received malformed msg: ${decodedBuffer}`,
      this.loggingMetadata,
    );
    return null;
  }

  if (!Value.Check(OpaqueTransportMessageSchema, parsedMsg)) {
    this.log?.error(`received invalid msg: ${JSON.stringify(parsedMsg)}`, {
      ...this.loggingMetadata,
      validationErrors: [
        ...Value.Errors(OpaqueTransportMessageSchema, parsedMsg),
      ],
    });

    return null;
  }

  return parsedMsg;
}
```

router/context.ts
```diff
@@ -41,7 +41,7 @@ export interface ServiceContext {}
  * ```
  */
 /* eslint-disable-next-line @typescript-eslint/no-empty-interface */
-export interface ParsedMetadata {}
+export interface ParsedMetadata extends Record<string, unknown> {}
```
masad-frost, line 44: good

transport/sessionStateMachine/common.ts
```diff
@@ -107,9 +107,14 @@ export interface SessionOptions {
...
   /**
-   * Duration to wait between connection disconnect and actual session disconnect
+   * Max duration that a session can be without a connection before we consider
+   * it dead.
    */
   sessionDisconnectGraceMs: number;
+  /**
+   * Whether to enable transparent session reconnects
+   */
+  enableTransparentSessionReconnects: boolean;
```
masad-frost, line 117: @zhounathan  FYI when this lands we should change AI chat to disable this explicitly instead of setting it to 0.

transport/sessionStateMachine/common.ts
```diff
@@ -107,9 +107,14 @@ export interface SessionOptions {
    */
   heartbeatsUntilDead: number;
   /**
-   * Duration to wait between connection disconnect and actual session disconnect
+   * Max duration that a session can be without a connection before we consider
+   * it dead.
```
masad-frost, line 111: Maybe a full on union is not worth it, but a comment is probably a good idea.

```suggestion
   * Max duration that a session can be without a connection before we consider
   * it dead. This value is ignored if {@link enableTransparentSessionReconnects} is
   * set to false.
```

Unless i'm misunderstanding something

transport/server.ts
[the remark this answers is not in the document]
masad-frost: i think it's a worthy future change to negotiate some of the options as part of the handshake, but for now it's on you to keep the options in sync as much as possible and deal with breaking option changes

lhchavez (approved):

transport/sessionStateMachine/transitions.ts
```diff
@@ -66,6 +69,7 @@ export const SessionStateGraph = {
         to,
         seq: 0,
         ack: 0,
+        graceExpiryTime: Date.now() + options.sessionDisconnectGraceMs,
```
masad-frost, line 72: nit opinion: I feel like this should be part of the  `SessionNoConnection` constructor.

transport/sessionStateMachine/common.ts
```diff
@@ -267,3 +272,43 @@ export abstract class IdentifiedSession extends CommonSession {
...
   }
 }
+
+export interface IdentifiedSessionWithGracePeriodListeners {
+  onSessionGracePeriodElapsed: () => void;
+}
+
+export interface IdentifiedSessionWithGracePeriodProps
+  extends IdentifiedSessionProps {
+  graceExpiryTime: number;
```
masad-frost, line 282: Talked in person about calling this something other than grace as the semantics of grace changed from a time-to-reconnect concern to a general time-to-connect concern

transport/sessionStateMachine/transitions.ts
```diff
@@ -112,11 +116,13 @@ export const SessionStateGraph = {
       listeners: SessionBackingOffListeners,
     ): SessionBackingOff => {
       const carriedState = inheritSharedSession(oldSession);
+      const graceExpiryTime = oldSession.graceExpiryTime;
```
masad-frost, line 119: nit: maybe we can do

```ts
function inheritSharedSessionWithGrace(
  session: IdentifiedSessionWithGracePeriod
): IdentifiedSessionWithGracePeriodProps {
  return {
    ...inheritSharedSession(session),
    graceExpiryTime: session.graceExpiryTime
  };
}
```

```suggestion
      const carriedState = inheritSharedSessionWithGrace(oldSession);
```

transport/sessionStateMachine/common.ts
masad-frost: Talked in person, looks like I missed the fact that session grace is now a general indicator for time-to-connect instead of time-to-reconnect

masad-frost (approved):

transport/sessionStateMachine/transitions.ts
jackyzha0: also talked in person, other states must pass this forward so its not always `Date.now() + options.sessionDisconnectGraceMs` 

e.g. `SessionBackingOff` -> `SessionNoConnection` must pass forward its grace expiry time

transport/sessionStateMachine/common.ts
jackyzha0: yup good idea i think the semantics of the option changed, i'll add a comment

PR #241, opened 2024-07-18
correctness, logging, and o11y fixes

jackyzha0:
## Why

- we had a few subtle bugs in
  - how we handled closing out of a state machine state
  - how we did logging
  - how we handled sessions during async breakpoints

## What Changed

See self-review comments for description

## Versioning

### DIFF
--- a/transport/client.ts
+++ b/transport/client.ts
@@ -317,6 +317,13 @@ export abstract class ClientTransport<
    * @param to The client ID of the node to connect to.
    */
   connect(to: TransportClientId) {
+    if (this.getStatus() !== 'open') {
+      this.log?.info(
+        `transport state is no longer open, cancelling attempt to connect to ${to}`,
+      );
+      return;
+    }
+
     // create a new session if one does not exist
     let session = this.sessions.get(to);
     session ??= this.createUnconnectedSession(to);
@@ -325,14 +332,6 @@ export abstract class ClientTransport<
       // already trying to connect
       this.log?.debug(
         `session to ${to} has state ${session.state}, skipping connect attempt`,
-        session.loggingMetadata,
-      );
-      return;
-    }
-
-    if (this.getStatus() !== 'open') {
-      this.log?.info(
-        `transport state is no longer open, cancelling attempt to connect to ${to}`,
         session.loggingMetadata,
       );
       return;
--- a/transport/client.ts
+++ b/transport/client.ts
@@ -247,13 +247,14 @@ export abstract class ClientTransport<
         : false;
 
       const reason = `handshake failed: ${msg.payload.status.reason}`;
+      const to = session.to;
       this.rejectHandshakeResponse(session, reason, {
         ...session.loggingMetadata,
         transportMessage: msg,
       });
 
       if (retriable) {
-        this.tryReconnecting(session.to);
+        this.tryReconnecting(to);
       } else {
         this.protocolError({
           type: ProtocolError.HandshakeFailed,
--- a/transport/client.ts
+++ b/transport/client.ts
@@ -255,7 +255,6 @@ export abstract class ClientTransport<
       if (retriable) {
         this.tryReconnecting(session.to);
       } else {
-        this.deleteSession(session);
         this.protocolError({
           type: ProtocolError.HandshakeFailed,
           code: msg.payload.status.code,
--- a/transport/impls/ws/client.ts
+++ b/transport/impls/ws/client.ts
@@ -64,9 +64,10 @@ export class WebSocketClientTransport extends ClientTransport<WebSocketConnectio
     });
 
     const conn = new WebSocketConnection(ws);
-    this.log?.info(`raw websocket to ${to} ok, starting handshake`, {
+    this.log?.info(`raw websocket to ${to} ok`, {
       clientId: this.clientId,
       connectedTo: to,
+      ...conn.loggingMetadata,
     });
 
     return conn;
--- a/transport/server.ts
+++ b/transport/server.ts
@@ -271,15 +271,69 @@ export abstract class ServerTransport<
     let oldSession = this.sessions.get(msg.from);
 
     // invariant: must pass custom validation if defined
-    const parsedMetadata = await this.validateHandshakeMetadata(
-      session,
-      oldSession,
-      msg.payload.metadata,
-      msg.from,
-    );
+    let parsedMetadata: ParsedMetadata = {};
+    if (this.handshakeExtensions) {
+      if (!Value.Check(this.handshakeExtensions.schema, msg.payload.metadata)) {
+        this.rejectHandshakeRequest(
+          session,
+          msg.from,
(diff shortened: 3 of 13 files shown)


transport/client.ts
```diff
@@ -247,15 +247,15 @@ export abstract class ClientTransport<
...
+      const to = session.to;
       this.rejectHandshakeResponse(session, reason, {
         ...session.loggingMetadata,
         transportMessage: msg,
       });
 
       if (retriable) {
-        this.tryReconnecting(session.to);
+        this.tryReconnecting(to);
       } else {
```
jackyzha0, line 258: `rejectHandshakeResponse` above already deletes the session, don't need to double-delete

transport/client.ts
```diff
@@ -318,6 +318,13 @@ export abstract class ClientTransport<
    * @param to The client ID of the node to connect to.
    */
   connect(to: TransportClientId) {
+    if (this.getStatus() !== 'open') {
+      this.log?.info(
+        `transport state is no longer open, cancelling attempt to connect to ${to}`,
+      );
+      return;
+    }
+
```
jackyzha0, line 327: check transport status before even trying to create a session

transport/impls/ws/client.ts
```diff
@@ -64,9 +64,10 @@ export class WebSocketClientTransport extends ClientTransport<WebSocketConnectio
     });
 
     const conn = new WebSocketConnection(ws);
-    this.log?.info(`raw websocket to ${to} ok, starting handshake`, {
+    this.log?.info(`raw websocket to ${to} ok`, {
       clientId: this.clientId,
       connectedTo: to,
+      ...conn.loggingMetadata,
```
jackyzha0, line 70: help track connections

transport/server.ts
```diff
@@ -271,15 +271,69 @@ export abstract class ServerTransport<
     let oldSession = this.sessions.get(msg.from);
 
     // invariant: must pass custom validation if defined
-    const parsedMetadata = await this.validateHandshakeMetadata(
-      session,
-      oldSession,
-      msg.payload.metadata,
-      msg.from,
-    );
+    let parsedMetadata: ParsedMetadata = {};
```
jackyzha0, line 274: inlining `validateHandshakeMetadata` so we can handle the async breakpoint

specifically, we should bail out if the session is consumed after the `await` point so we dont do stuff with a stale session

transport/sessionStateMachine/SessionConnecting.ts
```diff
@@ -59,7 +59,16 @@ export class SessionConnecting<
   // ends up rejected anyways
   bestEffortClose() {
     void this.connPromise
-      .then((conn) => conn.close())
+      .then((conn) => {
+        conn.close();
+        this.log?.info(
```
jackyzha0, line 64: better logging around this case

transport/sessionStateMachine/common.ts
```diff
@@ -70,6 +70,8 @@ abstract class StateMachineState {
         // modify _handleClose
         if (prop === '_handleClose') {
           return () => {
+            // target is the non-proxied object, we need to set _isConsumed again
+            target._isConsumed = true;
```
jackyzha0, line 74: lmao so `.close()` did not set `_isConsumed` before which caused some subtle bugs

transport/client.ts
```diff
@@ -247,15 +247,15 @@ export abstract class ClientTransport<
         : false;
 
       const reason = `handshake failed: ${msg.payload.status.reason}`;
+      const to = session.to;
```
masad-frost, line 250: did `session.to` change after reject or something?

transport/client.ts
```diff
@@ -318,6 +318,13 @@ export abstract class ClientTransport<
    * @param to The client ID of the node to connect to.
    */
   connect(to: TransportClientId) {
+    if (this.getStatus() !== 'open') {
```
masad-frost, line 321:
```suggestion
    if (this.getStatus() === 'closed') {
```
better glance value

masad-frost: should this happen? Feels like it should throw, things around it might hang if there's no feedback.

transport/server.ts
masad-frost: could we have done that inside the other function?

transport/server.ts
```diff
@@ -271,15 +271,69 @@ export abstract class ServerTransport<
...
+        : undefined;
+
+      const parsedMetadataOrFailureCode =
+        await this.handshakeExtensions.validate(
+          msg.payload.metadata,
+          previousParsedMetadata,
+        );
+
+      // double-check to make sure we haven't transitioned the session yet
+      if (session._isConsumed) {
```
masad-frost, line 308: feels wrong accessing a variable prefixed with `_`

transport/sessionStateMachine/SessionConnecting.ts
masad-frost: this will always end up logging, no?

transport/sessionStateMachine/common.ts
masad-frost: doesn't this end up calling `_handleStateExit` above?

transport/client.ts
jackyzha0: no we just dont allow reading anything from a stale session and reject deletes/closes the session so it threw

transport/transport.ts
```diff
@@ -209,8 +209,9 @@ export abstract class Transport<ConnType extends Connection> {
       session: session,
     });
 
+    const to = session.to;
```
masad-frost, line 212: maybe worth commenting why we're storing the variable before calling close?

masad-frost (approved):

transport/client.ts
masad-frost: right

transport/client.ts
jackyzha0: connect is just a connection attempt, its not guaranteed to establish a connection successfully by the time it returns

transport/server.ts
jackyzha0: we could have but imo the function didnt improve the readability that much, i got tripped up a few times remembering whether validateHandshakeMetadata did its own handshake rejection or not

transport/server.ts
jackyzha0: yeah :/ it was just so downstream consumers don't end up accessing it but i think its ok within library

transport/sessionStateMachine/SessionConnecting.ts
jackyzha0: bestEffort close is only called in the Connecting -> NoConnected case and the Connecting -> destroyed case

transport/sessionStateMachine/common.ts
jackyzha0: no that was the bug lol, target is the underlying object not the proxy

PR #243, opened 2024-07-24
deflake things, formalize protocol

jackyzha0:
## Why + What changed

- we rolled out `nextSentSeq` and `code` fully
- deflake `idle transport stays alive`

## Versioning


masad-frost (approved):
🧹

PR #244, opened 2024-07-25
make out of order non fatal + various fixes

jackyzha0:
## Why

- we see out-of-order messages in production but only from user agents `Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/126.0.0.0 Safari/537.36` and AS name of Menlo Security who seem like they run a proxy of some sort for security
- it is likely that their proxy drops websocket messages which sucks for us

## What changed

- no longer raise protocol error on out of order and instead close the connection
  - the reasoning is that the client will attempt to re-establish the connection and re-agree on the right seq number and resend the send buffer
  - if theres a disagreement at this stage, we will kill the session but otherwise we can actually transparently recover!
- also downgrade it to a warn
- extra check for session consumed during the async wait point when constructing the handshake

## Versioning

### DIFF
--- a/transport/events.ts
+++ b/transport/events.ts
@@ -8,6 +8,7 @@ export const ProtocolError = {
   RetriesExceeded: 'conn_retry_exceeded',
   HandshakeFailed: 'handshake_failed',
   MessageOrderingViolated: 'message_ordering_violated',
+  InvalidMessage: 'invalid_message',
 } as const;
 
 export type ProtocolErrorType =
--- a/__tests__/negative.test.ts
+++ b/__tests__/negative.test.ts
@@ -173,7 +173,7 @@ describe('should handle incompatabilities', async () => {
     });
   });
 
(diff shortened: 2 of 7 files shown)


transport/events.ts
```diff
@@ -8,6 +8,7 @@ export const ProtocolError = {
   RetriesExceeded: 'conn_retry_exceeded',
   HandshakeFailed: 'handshake_failed',
   MessageOrderingViolated: 'message_ordering_violated',
```
masad-frost, line 10: We're keeping this for the client?

masad-frost (approved):

transport/events.ts
jackyzha0: yep

jackyzha0: actually no its just there for backwards compat, should be safe to remove though

PR #245, opened 2024-07-25
label unhealthy sessions

jackyzha0:
## Why

- we'd like an easy way to filter session close events by whether the session ended up being a healthy or unhealthy session to get some baseline SLAs

## What changed

- add `unhealthy-session` tags to all the places where it makes sense

## Versioning

### DIFF
--- a/transport/server.ts
+++ b/transport/server.ts
@@ -508,7 +508,7 @@ export abstract class ServerTransport<
           onMessage: (msg) => this.handleMsg(msg),
           onInvalidMessage: (reason) => {
             this.protocolError({
-              type: ProtocolError.MessageOrderingViolated,
+              type: ProtocolError.InvalidMessage,
               message: reason,
             });
             this.deleteSession(connectedSession);
--- a/transport/server.ts
+++ b/transport/server.ts
@@ -16,7 +16,7 @@ import {
   ServerTransportOptions,
   defaultServerTransportOptions,
 } from './options';
-import { Transport } from './transport';
+import { DeleteSessionOptions, Transport } from './transport';
 import { coerceErrorString } from '../util/stringify';
 import { Static } from '@sinclair/typebox';
 import { Value } from '@sinclair/typebox/value';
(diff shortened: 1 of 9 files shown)


transport/server.ts
```diff
@@ -108,9 +108,12 @@ export abstract class ServerTransport<
     this.pendingSessions.delete(pendingSession);
   }
 
-  protected deleteSession(session: ServerSession<ConnType>): void {
+  protected deleteSession(
+    session: ServerSession<ConnType>,
+    unhealthy = false,
```
lhchavez, line 113: can we enum? `deletesession(foo, true)` doesn't tell me what `true` refers to. feels good but is bad.

transport/server.ts
```diff
@@ -511,7 +514,7 @@ export abstract class ServerTransport<
               type: ProtocolError.InvalidMessage,
               message: reason,
             });
-            this.deleteSession(connectedSession);
+            this.deleteSession(connectedSession, true);
```
lhchavez, line 517: is this triggered in the out-of-order case?

transport/transport.ts
```diff
@@ -201,8 +201,11 @@ export abstract class Transport<ConnType extends Connection> {
   }
 
   // state transitions
-  protected deleteSession(session: Session<ConnType>) {
-    session.log?.info(`closing session ${session.id}`, session.loggingMetadata);
+  protected deleteSession(session: Session<ConnType>, unhealthy = false) {
+    session.log?.info(`closing session ${session.id}`, {
+      ...session.loggingMetadata,
+      tags: unhealthy ? ['unhealthy-session'] : [],
```
lhchavez, line 207: do we need to concat this with `session.loggingMetadata.tags ?? []`?

lhchavez (approved):

transport/transport.ts
jackyzha0: ah yes

transport/server.ts
jackyzha0: not anymore, this is just failure to parse a message (we no longer delete a session on out-of-order as of https://github.com/replit/river/pull/244)

transport/server.ts
masad-frost: Or "options" object

PR #247, opened 2024-07-26
close with 1000 instead of 1005

jackyzha0:
## Why

- its healthy :)

## What changed

- change ws close to close with 1000 going away instead of 1005 no close code

## Versioning

### DIFF
--- a/transport/impls/ws/connection.ts
+++ b/transport/impls/ws/connection.ts
@@ -5,6 +5,8 @@ interface ConnectionInfoExtras {
   headers: Record<string, string>;
 }
 
+const WS_HEALTHY_CLOSE_CODE = 1000;
+
 export class WebSocketConnection extends Connection {
   ws: WsLike;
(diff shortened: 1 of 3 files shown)


transport/impls/ws/connection.ts
```diff
@@ -65,6 +65,10 @@ export class WebSocketConnection extends Connection {
   }
 
   close() {
-    this.ws.close();
+    // 1001 is going away, this is a 'healthy' close
+    // even if its not really healthy at an application level
+    // if we don't specify this, it defaults to 1005 which
+    // some proxies/loggers detect as an error
+    this.ws.close(1001);
```
lhchavez, line 72: why not 1000? 1001 is more for things like "this server is shutting down", and 1000 is more for "kthxbye, i have nothing more to discuss"

lhchavez (approved):
lgtm with the change of code. bonus points if we declare a named constant for this.

PR #248, opened 2024-07-30
[protocolv2] Remove the concept of close requests

masad-frost:
## Why

Came up as part of the giga review for #127.

It's just one more concept that's unneeded. Instead of close requests, streams can listen to the closes on the other pipe and decided if they want to close then. Alternatively (for upload and stream procs), service implementers can implement a user-space close-request mechanisms. The only thing that this leaves open is in subscriptions clients have no way of signaling disinterest in the data, in which case they should just abort the stream.

FWIW this is how gRPC works.

## What changed

Remove close request, swap with aborts in tests where relevant.


Monkatraz (approved):
lgtm, seems easier to think about this way

jackyzha0:
i guess if the subscription needs to be cancellable like that, should be ok to structure it as a stream anyways

jackyzha0 (approved):

masad-frost:
closing in favor of #249

PR #249, opened 2024-07-31
RFC: take 2 on streams

masad-frost:
Got some comments on streams being complicated, so this is a take on simplified API.

- Less methods and states that people need to understand
- Merge interfaces into a single object (i.e. a `stream` call would have both read and write capabilities on the same object)


router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+      next: async () => {
+        if (this.broken && didSignalDrain) {
+          return {
+            done: true,
+            value: undefined,
+          } as const;
+        }
+
+        // Wait until we have something in the queue
+        while (this.queue.length === 0) {
```
jackyzha0, line 153: is having while here a mistake? seems like we dont modify `this.queue` in the loop?

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+  /**
+   * Whether break was called.
+   */
+  private broken = false;
+  /**
+   * This flag allows us to avoid cases where break was called,
+   * but the stream is fully consumed and closed. We don't need
+   * to signal that break was called, only that the stream is closed.
+   */
+  private didBreakDisposeValues = false;
```
jackyzha0, line 119: i think this could be clearer as the non-negative `brokenWithValuesLeftToRead` or something in that vein

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+   */
+  private resolveNextPromise: null | (() => void) = null;
+
+  public [Symbol.asyncIterator]() {
+    if (this.locked) {
+      throw new TypeError('Readable is already locked');
+    }
+
+    // first iteration with break signals an error, the following one signals end of iteration.
+    let didSignalDrain = false;
```
jackyzha0, line 140:
```suggestion
    let didSignalBreak = false;
```

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+        }
+
+        // Unfortunately we have to use non-null assertion here, because T can be undefined
+        // we already check for array length above anyway
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const value = this.queue.shift()!;
+
+        return { done: false, value } as const;
+      },
+      return: () => {
```
jackyzha0, line 188: wut is this a part of the async iterator api?

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+   * Used by methods in the class to signal to the iterator that it
+   * should check for the next value.
+   */
+  private nextPromise: Promise<void> | null = null;
+  /**
+   * Resolves nextPromise
+   */
+  private resolveNextPromise: null | (() => void) = null;
+
+  public [Symbol.asyncIterator]() {
```
jackyzha0, line 134: i wonder if we can actually just model this as an async generator and avoid `nextPromise` entirely

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator#instance_methods

because it inherits asynciterator we can keep the return type the same but the internal implementation is a generator

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+    this.locked = true;
+    this.broken = true;
+    this.didBreakDisposeValues = this.queue.length > 0;
+    this.queue.length = 0;
+
+    // if we have any iterators waiting for the next value,
+    this.resolveNextPromise?.();
+  }
+
+  public isReadable(): boolean {
```
jackyzha0, line 218:
```suggestion
  get readable(): boolean {
```

nit: in my mind readable almost always feels like a prop

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    this.closed = true;
+    this.resolveNextPromise?.();
+  }
+
+  /**
+   * @internal meant for use within river, not exposed as a public API
+   */
+  public _hasValuesInQueue(): boolean {
```
jackyzha0, line 258: doesn't seem like we use this

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+export interface Writable<T> {
+  /**
+   * {@link write} writes a value to the stream. An error is thrown if writing to a closed stream.
+   */
+  write(value: T): undefined;
+  /**
+   * {@link close} signals the closure of the write stream, informing the reader that the stream is complete.
+   * Calling {@link close} multiple times has no effect.
+   */
+  close(): undefined;
```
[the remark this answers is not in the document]
masad-frost: Yeah, I went back and forth and settled on `close`. Partly because that's the word we use in our protocol, and it's what web streams settled on.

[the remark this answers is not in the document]
masad-frost: What is it about "end" that you like?

router/streams2.ts
```diff
@@ -0,0 +1,310 @@
...
+      throw new TypeError('Readable is already locked');
+    }
+
+    // first iteration with break signals an error, the following one signals end of iteration.
+    let didSignalDrain = false;
+    this.locked = true;
+
+    return {
+      next: async () => {
+        if (this.broken && didSignalDrain) {
```
masad-frost, line 145:
```suggestion
        if (didSignalDrain) {
```

router/streams2.ts
masad-frost: We don't modify it directly, but we wait for it to be modified. The block will run only once if there are values pushed, and will run twice if we get into a broken state or closed state.

Can probably be structured better, open to suggestions.

router/streams2.ts
masad-frost: I'm not a fan of getters at all, I think they're terrible and should only be used when you have to use them (i.e. conforming to an interface that you have no control over). They're implicit function calls that obscure intention and inhibit the reader from knowing what is actually happening or what's possible.

router/streams2.ts
masad-frost: we use it in tests pretty sure

router/streams2.ts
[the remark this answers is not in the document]
jackyzha0: https://nodejs.org/api/stream.html#writableendchunk-encoding-callback node.js uses end in their writable so it just feels natural having spent time in that ecosystem

router/procedures.ts
```diff
@@ -129,13 +137,44 @@ export interface RpcProcedure<
...
-  }: {
-    ctx: ProcedureHandlerContext<State>;
-    reqInit: Static<RequestInit>;
-  }): Promise<Result<Static<ResponseData>, Static<ResponseErr>>>;
+  handler(
+    param: BaseProcedureHandlerParam<State, RequestInit>,
+  ): Promise<Result<Static<ResponseData>, Static<ResponseErr>>>;
+}
+
+export class UploadProcedureHandlerParam<
```
masad-frost, line 145: @jackyzha0 any comments on this thingy?

jackyzha0: no this seems fine to me

router/streams2.ts
masad-frost: yeah, it's called when `break` or `return` happens. Part of the reason I wanna expose only a simple iterator is that I don't want people messing with iteration via `iter.throw` or `iter.return`

router/streams2.ts
jackyzha0: see comment below re: generators, could be interesting to try

router/streams2.ts
jackyzha0: yeah fair

router/procedures.ts
masad-frost: sweet, i'll address comments and finish the implementation (I'm mostly done)

router/streams2.ts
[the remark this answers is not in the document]
masad-frost: I see, web on the other hand uses "close" :cry: https://developer.mozilla.org/en-US/docs/Web/API/WritableStream/close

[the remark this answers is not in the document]
jackyzha0: ripo yeah seems like theres disagreement, im ok keeping the web interface

router/procedures.ts
masad-frost: Do you think we should put `abort` here instead of on the context?

masad-frost: I don't feel like I have a clear idea of what should go into ctx. Maybe anything but the data is a good framework?

router/streams2.ts
Monkatraz: I made a very similar class to this for a personal project and had to use this exact same pattern, actually not sure if you can use a generator here

router/streams2.ts
Monkatraz: yeah not sure if this needs to be a `while`, couldn't you just check if there are any values in the queue or not?

router/streams2.ts
masad-frost: maybe, i remember trying it and not working or ending up with the same boilerplate but with less control.

router/streams2.ts
masad-frost: yeah we can check that once, wait for the promise then do the same checks again. I'm pretty sure it can be structured as 3 if statements with the same body.

masad-frost: i'll add a comment for now, at the very least people reading this won't be as confused.

masad-frost: For science, I tried using if statements and thankfully ran into a failing test that had it covered. In a normal iteration case the while loop can be structured as a couple of if and the loop will run at most a couple of times:
 - it will run 0 times if we have something in the queue to consume
 - it will run 1 time if we have nothing in the queue and then get something in the queue
 - it will run 2 times if we have nothing in the queue and then the readable closes or breaks
 
However, in a degenerate case where something has the handle to the iterator and is calling `next` eagerly multiple times this loop will come in handy by queuing them up.

masad-frost: we could set up a queuing mechanism that is more explicit, but i'll just add a comment unless one of you is volunteering to take care of it.

router/procedures.ts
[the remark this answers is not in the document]
masad-frost: How about `UploadReadable`? I'm trying to avoid using the term `stream` tbh, it's quite overloaded

router/client.ts
```diff
@@ -72,8 +73,7 @@ type UploadFn<
 > = (
   reqInit: ProcInit<Router, ProcName>,
   options?: CallOptions,
-) => {
-  reqWriter: WriteStream<ProcInput<Router, ProcName>>;
+) => Writable<ProcInput<Router, ProcName>> & {
```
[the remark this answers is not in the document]
masad-frost: will fix in another PR when fixing the client.

masad-frost:
As much as I hate to merge into protocolv2 and leaving the branch non-functional/broken, I'm gonna go ahead an merge this to keep my PRs small and build on top of these abstractions without dealing with stacked branches too much.

PR #250, opened 2024-08-01
[protocolv2] Use new simpler readable and writable interfaces

masad-frost:
## Why

Follow up on #249 to actually use the new interfaces. Removed extra refactor done in #249 that merges the interfaces as that proved to be a challenging API (un-yak-shave :no_good: :water_buffalo:)

## What changed

- Removed close requests (mostly cherry-picked from #248)
- Otherwise a simple swapping out of the interfaces

## Versioning


masad-frost:
I'll be merging this since it's a simple switch-aroo, review at your own pace. It doesn't address any of the comments on the base PR, so there will be other follow ups. I'll also rename "reader" and "writer" in a separate PR.

PR #254, opened 2024-08-14
Introduce $kind as a pivot point for uncaught errors

blast-hardcheese:
## Why

`$kind` is emerging as a de-facto property to identify the type of a payload. Adding a literal `$kind: "error"` permits consumers to easily filter in/out error payloads.

## What changed

Add `$kind: "error"` to `RiverUncaughtSchema`, for consumers to easily identify error payloads

## Versioning

### DIFF
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -333,6 +333,7 @@ describe.each(testMatrix())(
       expect(result3).toStrictEqual({
         ok: false,
         payload: {
+          $kind: 'error',
           code: UNCAUGHT_ERROR,
           message: 'some message',
         },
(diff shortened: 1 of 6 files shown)


jackyzha0 (changes requested):
`code` already exists for this purpose, all new user defined errors also define their own `code` to discriminate on

https://github.com/replit/river/blob/main/router/result.ts#L14

blast-hardcheese:
Classifiers are useful, since `code` idiomatically means the actual value to match on, not that the structure represents an error more broadly.

jackyzha0:
if its `ok: false` then its an error, is there anything else in particular you're trying to assert?

blast-hardcheese:
Ahh. The issue I was running into was that the python codegen doesn't expose `ok`, and the results of the client operations are just an otherwise inscrutable `R | E`. Bummer.

blast-hardcheese:
Alright. Going with a different approach.

PR #255, opened 2024-08-16
Fix closing readable twice sometimes on session disconnect

masad-frost:
## Why

Seeing this case sometimes where:
1. Stream created
2. Server closes
3. Session disconnect
4. Leads to an uncaught error

There aren't any real side-effects from this AFAICT, but it's a bug

## What changed

- Fixed the issue, only close the readable if it hasn't been closed already
- Added a check elsewhere that doesn't have a guard around closeRedable

## Versioning

### DIFF
--- a/router/client.ts
+++ b/router/client.ts
@@ -443,7 +443,13 @@ function handleProc(
     if (isStreamClose(msg.controlFlags)) {
       span.addEvent('received response close');
 
-      closeReadable();
+      if (resReadable.isClosed()) {
+        transport.log?.error(
+          'received stream close but readable was already closed',
+        );
+      } else {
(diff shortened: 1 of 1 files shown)


router/client.ts
```diff
@@ -443,7 +443,13 @@ function handleProc(
...
       span.addEvent('received response close');
 
-      closeReadable();
+      if (resReadable.isClosed()) {
+        transport.log?.error(
+          'received stream close but readable was already closed',
+        );
+      } else {
+        closeReadable();
+      }
```
masad-frost, line 452: never ran into this, but thought i'd log

jackyzha0 (approved):

PR #259, opened 2024-08-20
silence dts

jackyzha0:
## Why

```
Export "ClientTransport" of module "transport/client.ts" was reexported through module "transport/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
Either change the import in "util/testHelpers.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
Export "ServerTransport" of module "transport/server.ts" was reexported through module "transport/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
Either change the import in "util/testHelpers.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
Export "ClientTransport" of module "transport/client.ts" was reexported through module "transport/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
Either change the import in "util/testHelpers.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
Export "ServerTransport" of module "transport/server.ts" was reexported through module "transport/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
Either change the import in "util/testHelpers.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
```

## Versioning

### DIFF
--- a/util/testHelpers.ts
+++ b/util/testHelpers.ts
@@ -30,13 +30,14 @@ import {
   Session,
   SessionStateGraph,
 } from '../transport/sessionStateMachine/transitions';
-import { ClientTransport, ServerTransport } from '../transport';
 import {
   BaseErrorSchemaType,
   ProcedureErrorSchemaType,
   ReaderErrorSchema,
   UNCAUGHT_ERROR_CODE,
 } from '../router/errors';
+import { ClientTransport } from '../transport/client';
+import { ServerTransport } from '../transport/server';
 
 /**
  * Creates a WebSocket client that connects to a local server at the specified port.


masad-frost (approved):
We had many more instances of circular imports last I checked, I had a branch that deals with them, quite outdated at this point, but we can revisit later

PR #260, opened 2024-08-22
Fix ctx.signal documentation

masad-frost:
## Why

It didn't make sense, and wasn't clear.

## What changed

Update docs

Maybe worth considering renaming to "invocationFinishedSignal" or something more verbose, as signals in JavaScript are typically used in context of cancellation.

## Versioning

### DIFF
--- a/router/context.ts
+++ b/router/context.ts
@@ -73,7 +73,12 @@ export type ProcedureHandlerContext<State> = ServiceContext & {
   /**
    * This signal is a standard [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal)
    * triggered when the procedure invocation is done. This signal tracks the invocation/request finishing
-   * for _any_ reason including the procedure invocation.
+   * for _any_ reason, for example:
+   * - client explicit cancellation
+   * - procedure handler explicit cancellation via {@link cancel}
+   * - client session disconnect
+   * - server cancellation due to client invalid payload
+   * - invocation finishing cleanly (i.e. rpc handler return, or both sides closing streams)
    *
    * You can use this to pass it on to asynchronous operations (such as fetch).
    *


Monkatraz (approved):
love better docs

I think being more verbose with the name is fine as long as you don't have to pass this signal to a bunch of things - if you constantly have to use it, I think a more terse name is fine

router/context.ts
```diff
@@ -73,7 +73,12 @@ export type ProcedureHandlerContext<State> = ServiceContext & {
   /**
    * This signal is a standard [AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal)
    * triggered when the procedure invocation is done. This signal tracks the invocation/request finishing
-   * for _any_ reason including the procedure invocation.
+   * for _any_ reason, for example:
+   * - client explicit cancellation
+   * - procedure handler explicit cancellation via {@link cancel}
+   * - client session disconnect
+   * - server cancellation due to client invalid payload
+   * - invocation finishing cleanly (i.e. rpc handler return, or both sides closing streams)
```
lhchavez, line 81:
```suggestion
   * - invocation finishing cleanly (i.e. rpc handler return, or either sides closing streams)
```

from my understanding, the signal happens when either side closes the stream. if it doesn't, might be worth adding a clarification.

lhchavez (approved):

router/context.ts
masad-frost: No, it's both closing their end. Half-close doesn't trigger the signal. I'll add clarification.

jackyzha0: maybe this?

`invocation finishing cleanly (i.e. rpc handler return, or after the client-side has ended the writable and the server-side has ended the readable)`

jackyzha0 (approved):

router/context.ts
masad-frost, line 81:
```suggestion
   * - invocation finishes cleanly, this depends on the type of the procedure (i.e. rpc handler return, or in a stream after the client-side has closed the request writable and the server-side has closed the response writable)
```

PR #261, opened 2024-08-24
Use snapshot testing for schema serialization

masad-frost:
## Why

It's already kind of snapshot test via copy-paste. Cumbersome to update via copy pasting, let's use the tool!

## What changed

Use snapshot testing for schema serialization

## Versioning

- [ ] Breaking protocol change
- [ ] Breaking ts/js API change

PR #262, opened 2024-08-24
Disallow union of code in BaseErrorSchema

masad-frost:
## Why

Currently we allow a union of error objects and those objects can have a union of codes. For consistency, especially to simplify codegen, let's just expect a union of objects. 

## What changed

|  image |  unionofcode |
|---|---|
|  image | unionoferror  |

## Versioning

### DIFF
--- a/router/errors.ts
+++ b/router/errors.ts
@@ -32,11 +32,11 @@ type TLiteralString = TLiteral<string>;
 
 export type BaseErrorSchemaType =
   | TObject<{
-      code: TLiteralString | TUnion<Array<TLiteralString>>;
+      code: TLiteralString;
       message: TLiteralString | TString;
     }>
   | TObject<{
-      code: TLiteralString | TUnion<Array<TLiteralString>>;
+      code: TLiteralString;
       message: TLiteralString | TString;
       extras: TSchema;
     }>;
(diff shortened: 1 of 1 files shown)


masad-frost:

* main:
  * **PR #261**
    * **PR #262** 👈
      * **PR #263**

router/errors.ts
```diff
@@ -32,19 +32,19 @@ type TLiteralString = TLiteral<string>;
 
 export type BaseErrorSchemaType =
   | TObject<{
-      code: TLiteralString | TUnion<Array<TLiteralString>>;
+      code: TLiteralString;
       message: TLiteralString | TString;
     }>
   | TObject<{
-      code: TLiteralString | TUnion<Array<TLiteralString>>;
+      code: TLiteralString;
```
lhchavez, line 39: is there an advantage of having this be a union of two types that only differ by an optional `extras`?

this is giving me "corporate wants you to find the differences between these two pictures" vibes.

lhchavez (approved):

lhchavez (approved):
also, is this the new meta?

no-more-drake-react-the-new-wave-is-not-like-us-react-v0-l7a7ridqlqad1

router/errors.ts
masad-frost: Honestly we haven't used extras for now, maybe a YAGNI and we should get rid of it for now.

masad-frost: But to answer your question, yes this is necessary. You can't express optional schema fields

image

masad-frost: https://replit.com/@replitfaris/SizzlingVictoriousMeasurements#index.ts

jackyzha0 (approved):

PR #263, opened 2024-08-24
Include protocol defined errors in schema serialization

masad-frost:
## Why

The "built-in" errors can be emitted by any procedure, we should include them in our serialized format so that we don't have to merge them in everywhere that consumes serialized schema for codegen

## What changed

All serialized procedures now include protocol errors

## Versioning

### DIFF
--- a/router/services.ts
+++ b/router/services.ts
@@ -1,4 +1,4 @@
-import { Type, TSchema, Static } from '@sinclair/typebox';
+import { Type, TSchema, Static, Kind, TNever } from '@sinclair/typebox';
 import {
   Branded,
   ProcedureMap,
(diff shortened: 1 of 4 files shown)


masad-frost:

* main:
  * **PR #261**
    * **PR #262**
      * **PR #263** 👈

router/services.ts
```diff
@@ -431,16 +435,11 @@ export class ServiceSchema<
           {
             init: Type.Strict(procDef.requestInit),
             output: Type.Strict(procDef.responseData),
+            errors: getSerializedProcErrors(procDef),
```
jackyzha0, line 438: nice

jackyzha0 (approved):

PR #264, opened 2024-08-31
log uncaught errors in handlers

jackyzha0:
## Why

having o11y into why procedures died unexpectedly is good, we used to have this but lost it in the migration

## What changed

re-log errors in procs

## Versioning

### DIFF
--- a/router/server.ts
+++ b/router/server.ts
@@ -509,6 +509,10 @@ class RiverServer<Services extends AnyServiceSchemaMap>
         `${serviceName}.${procedureName} handler threw an uncaught error`,
         {
           ...loggingMetadata,
+          transportMessage: {
+            procedureName,
+            serviceName,
+          },
           extras: {
             error: errorMsg,
           },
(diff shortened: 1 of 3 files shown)


JimmyAustin (approved):

router/server.ts
```diff
@@ -518,10 +528,7 @@ class RiverServer<Services extends AnyServiceSchemaMap>
     } else if (procedure.type === 'rpc' || procedure.type === 'subscription') {
       // Though things can work just fine if they eventually follow up with a stream
       // control message with a close bit set, it's an unusual client implementation!
-      this.log?.warn('sent an init without a stream close', {
-        ...loggingMetadata,
-        clientId: this.transport.clientId,
-      });
+      this.log?.warn('sent an init without a stream close', loggingMetadata);
```
lhchavez, line 531: does the logging metadata already have the client id? or why sre we yeeting this?

lhchavez (approved):

router/server.ts
jackyzha0: ye it does, this was an extraneous spread

PR #266, opened 2024-09-03
Add a special logging tag to uncaught handler exceptions

masad-frost:
## Why

We want to pipe uncaught errors in handlers to Sentry, we think it's ok to do that via the logger as opposed to having a special error callback.

## What changed

- Add tag `uncaught-handler-error`
- Include `originalException` in `extras` (for proper stack traces)
- Change `extras` in logs from `unknown` to an optional `Record<string, unknown>` (we're using it that way anyway, just makes it easier to handle)

## Versioning

### DIFF
--- a/transport/impls/ws/connection.ts
+++ b/transport/impls/ws/connection.ts
@@ -1,8 +1,8 @@
 import { Connection } from '../../connection';
 import { WsLike } from './wslike';
 
-interface ConnectionInfoExtras {
-  headers: Record<string, string>;
+interface ConnectionInfoExtras extends Record<string, unknown> {
+  headers: Record<string, unknown>;
 }
 
 const WS_HEALTHY_CLOSE_CODE = 1000;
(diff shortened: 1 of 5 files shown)


transport/impls/ws/connection.ts
```diff
@@ -1,8 +1,8 @@
 import { Connection } from '../../connection';
 import { WsLike } from './wslike';
 
-interface ConnectionInfoExtras {
-  headers: Record<string, string>;
+interface ConnectionInfoExtras extends Record<string, unknown> {
+  headers: Record<string, unknown>;
```
jackyzha0, line 5: can we keep this string

jackyzha0 (approved):

transport/impls/ws/connection.ts
Monkatraz, line 5: are headers not always `Record<string, string>`?

Monkatraz (approved):
lgtm

transport/impls/ws/connection.ts
masad-frost: oh sure.

masad-frost: looks like jacky transforms them!

PR #270, opened 2024-09-14
re-add async dispose to services

jackyzha0:
## Why

cleaning up services in tests is helpful, esp if we do things like open ports

## What changed

add `.close` to the server router

## Versioning

### DIFF
--- a/router/server.ts
+++ b/router/server.ts
@@ -70,6 +70,8 @@ export interface Server<Services extends AnyServiceSchemaMap> {
    * A set of stream ids that are currently open.
    */
   streams: Map<StreamId, ProcStream>;
+
+  close: () => Promise<void>;
 }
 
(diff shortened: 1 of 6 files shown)


router/server.ts
```diff
@@ -912,6 +919,15 @@ class RiverServer<Services extends AnyServiceSchemaMap>
     const msg = cancelMessage(streamId, payload);
     sessionScopedSend(msg);
   }
+
+  async close() {
+    this.unregisterTransportListeners();
+
+    for (const serviceName of Object.keys(this.services)) {
+      const service = this.services[serviceName];
+      await service[Symbol.asyncDispose]();
```
masad-frost, line 928: try catch

masad-frost (approved):

PR #272, opened 2024-09-27
Log malformed message contents in base64

bradymadden97:
## Why

Noticed a number of malformed message contents in DD https://us5.datadoghq.com/logs?query=%22received%20malformed%20msg%22%20%40environment%3Aproduction&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice%2C%40riverTarget&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1726796110823&to_ts=1727400910823&live=true

However we are sending binary messages originally, so we can't tell what the original contents were once its decoded to a string. We can base64 encode the buffer and then reverse it when trying to debug what is going wrong with these messages 

## Versioning

### DIFF
--- a/transport/sessionStateMachine/common.ts
+++ b/transport/sessionStateMachine/common.ts
@@ -160,9 +160,8 @@ export abstract class CommonSession extends StateMachineState {
     const parsedMsg = this.options.codec.fromBuffer(msg);
 
     if (parsedMsg === null) {
-      const decodedBuffer = new TextDecoder().decode(Buffer.from(msg));
       this.log?.error(
-        `received malformed msg: ${decodedBuffer}`,
+        `received malformed msg: ${Buffer.from(msg).toString('base64')}`,
         this.loggingMetadata,
       );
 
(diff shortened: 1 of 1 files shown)


jackyzha0 (approved):
b64 is nice, we have validation for the schema though, do we have enough data from that?

transport/sessionStateMachine/common.ts
```diff
@@ -160,22 +160,24 @@ export abstract class CommonSession extends StateMachineState {
...
-      });
+      this.log?.error(
+        `received invalid msg: ${Buffer.from(msg).toString('base64')}`,
+        {
+          ...this.loggingMetadata,
+          validationErrors: [
+            ...Value.Errors(OpaqueTransportMessageSchema, parsedMsg),
+          ],
+        },
+      );
```
jackyzha0, line 180: actually prefer this one as is, we have a parsed msg here already

jackyzha0: doesnt make sense as b64 idt

bradymadden97: sounds good will revert this one

PR #275, opened 2024-10-16
make ReadableIterator an async iterable iterator

jackyzha0:
## Why

sometimes nice to be able to do `.next` and still be able to use the iterator in an `for await ... of` loop

## What changed

implement the trivial `Symbol.asyncIterator`

this still retains 'single reader' semantics, if you have two asyncIterators from a single readable, they will just compete for the next value (same as if you had two methods both calling `.next` on the resulting iterator)

## Versioning


jackyzha0:
we decided to keep the api surface cleaner, you can do this with a check inside the `for await ... of` loop

PR #276, opened 2024-10-18
make handler spans more accurate, re-add connection span

jackyzha0:
## Why

- a lot of proc handler specific log messages didnt get picked up by the right tracer
- move proc stream <> handler proxy messages into the handler span
- move the invoked message into the client handler span
- re-add connection telemetry
- span linking between procs/handlers and their originating sessions
- we use https://www.npmjs.com/package/@opentelemetry/context-async-hooks so context is tracked safely in an async context so we should leverage that

## Versioning

### DIFF
--- a/logging/log.ts
+++ b/logging/log.ts
@@ -1,5 +1,6 @@
 import { ValueError } from '@sinclair/typebox/value';
 import { OpaqueTransportMessage, ProtocolVersion } from '../transport/message';
+import { context, trace } from '@opentelemetry/api';
 
 const LoggingLevels = {
   debug: -1,
@@ -27,6 +28,17 @@ export type Tags =
 
 const cleanedLogFn = (log: LogFn) => {
   return (msg: string, metadata?: MessageMetadata) => {
+    // try to infer telemetry
+    if (metadata && !metadata.telemetry) {
+      const span = trace.getSpan(context.active());
+      if (span) {
+        metadata.telemetry = {
+          traceId: span.spanContext().traceId,
+          spanId: span.spanContext().spanId,
+        };
+      }
+    }
+
     // skip cloning object if metadata has no transportMessage
     if (!metadata?.transportMessage) {
       log(msg, metadata);
@@ -37,6 +49,7 @@ const cleanedLogFn = (log: LogFn) => {
     // clone metadata and clean transportMessage
     const { payload, ...rest } = metadata.transportMessage;
     metadata.transportMessage = rest;
+
     log(msg, metadata);
   };
 };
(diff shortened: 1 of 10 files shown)


logging/log.ts
```diff
@@ -27,6 +28,17 @@ export type Tags =
 
 const cleanedLogFn = (log: LogFn) => {
   return (msg: string, metadata?: MessageMetadata) => {
+    // try to infer telemetry
+    if (metadata && !metadata.telemetry) {
+      const span = trace.getSpan(context.active());
+      if (span) {
+        metadata.telemetry = {
```
blast-hardcheese, line 35: I would not expect an innocuous logging function to modify the metadata here. How are these metadata created? Can we just be more strict instead of trying to heal post-hoc? Removing optionality increases intelligibility

blast-hardcheese (approved):
Looks sensible, one comment

logging/log.ts
jackyzha0: this metadata are manually injected at logging callsites, theres a lot and we kept finding places where we didnt inject the right telemetry :( 

i thought it would be easier to 'do the right thing' and follow the open telemetry suggestion of just always logging the active telemetry

this doesnt log anything if its not within an active span

jackyzha0: it also means that any logging inside a user-written handler also gets the right span associations

blast-hardcheese: I still wouldn't expect it, is the concern that they're too expensive to preallocate or what?

blast-hardcheese: Yeah, the underlying concern is the implicit complexity. In a logging method, we take an optional parameter with an optional member, then we write that field before returning back.

A logging function definitely doesn't seem to be the right place to be doing any of that work.

Ultimately it's not the end of the world if we go this way, but I'd much rather solve this by making it easier to create and manage the `MessageMetadata` instead of offering important functionality for free invisibly just by calling `log`. If that's easy, then the span fields could even be marked required and defensive complexity can be delegated to where it belongs.

lhchavez: of note, this is exactly how most third-party logging libraries operate: they add a global hook where they perform this same operation of massaging the metadata, only to avoid callers having to reason about adding metadata explicitly so that _all_ logs get the tracing metadata.

lhchavez (approved):

router/server.ts
```diff
@@ -263,6 +267,12 @@ class RiverServer<Services extends AnyServiceSchemaMap>
       id: sessionId,
     } = initialSession;
 
+    // dont use the session span here, we want to create a new span for the procedure
```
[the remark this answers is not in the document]
jackyzha0: yes! i can add a comment but we add the link already in span itself, see `createProcTelemetryInfo`

PR #277, opened 2024-10-22
properly end connection span

jackyzha0:
## Why

we weren't ending the connection span before!

## What changed

end the connection span when the connection ends

## Versioning


blast-hardcheese (approved):
nice

PR #278, opened 2024-10-23
Encode messages before storing in send buffer

cbrewster:
## Why

I want to start enforcing max payload size before we attempt to send the messages over the wire.

Ideally we can confine these errors to be per-stream so one bad RPC call doesn't explode the entire client.

As a pre-req, we need to check for the max payload size before accepting messages into the send buffer, which means we need to encode them first to figure out the full payload size.

An additional benefit is we only ever encode messages once rather than re-encoding every time we resend messages from the send buffer. Granted, this is probably not a path we hit very often.

## What changed

- Encode messages before storing them in the send buffer
    - We keep around the id / seq number since we do need to have access to these for some tests and for keeping track of which messages have been acked by the peer and can be dropped from the send buffer

## Versioning

### DIFF
--- a/testUtil/fixtures/cleanup.ts
+++ b/testUtil/fixtures/cleanup.ts
@@ -1,4 +1,4 @@
-import { expect, vi } from 'vitest';
+import { assert, expect, vi } from 'vitest';
 import {
   ClientTransport,
   Connection,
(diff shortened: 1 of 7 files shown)


cbrewster:

* main:
  * **PR #278** 👈

jackyzha0 (approved):

jackyzha0:
was done in https://github.com/replit/river/pull/363

PR #279, opened 2024-10-24
Enforce maximum payload size

cbrewster:
## Why

Memory is not unbounded and underlying transports have limits on maximum payload size. When we accidentally exceed the max payload size for an underlying transport, the connection can get torn down entirely. This is bad because it means one bad RPC call can bring down the entire connection.

Additionally, when things do explode, its hard to find the right error message that tells you that this was due to exceeding some payload limit.

Let's enforce a configurable max payload limit in river itself. It's up to users to make sure that this max payload size is `<=` the max payload size allowed by their transport layer.

Something to consider in the future is automatically chunking these messages in river; however, I think this is a bad idea because we want to avoid having large buffers in memory entirely. Having this restriction in river will force application developers to stream or chunk data and help avoid cases where loading large buffers into memory happens to work until we hit OOMs.

## What changed

- Add config value for setting the max payload size
    - Default is 4M to match gRPC
- Add new built-in error code: `MAX_PAYLOAD_SIZE_EXCEEDED`
- If we exceed the max payload size after encoding, throw and error
- Handle these new errors accordingly in the client procedure handler
    - If init message: inject error in the readable only client side. Server never sees these.
    - If in the middle of a stream or upload: cancel the request on the server, inject error in the readable, throw error since `write` returns `void`
- Handle these new errors accordingly in the server procedure handler
    - Return error with code `MAX_PAYLOAD_SIZE_EXCEEDED` whenever server attempts to send a payload that is too large
- Updated the protocol docs to reflect this new error

## Versioning

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -162,7 +162,7 @@ All messages MUST have no control flags set (i.e., the `controlFlags` field is `
   - The payload MUST be `{ type: 'ACK' }`.
   - Because this is a control message that is not associated with a specific stream, you MUST NOT set `serviceName` or `procedureName` and `streamId` can be something arbitrary (e.g. `heartbeat`).
 
-There are 4 error payloads that are defined in the protocol sent from server to client, these codes are reserved:
+There are 5 error payloads that are defined in the protocol sent from server to client, these codes are reserved:
 
 ```ts
 // When a client sends a malformed request. This can be
@@ -187,11 +187,22 @@ interface CancelError extends BaseError {
   message: string;
 }
 
+// This occurrs when a client or server attempts to send a message that is
+// larger than the configurable max payload size.
+interface MaxPayloadSizeExceededError extends BaseError {
+  code: 'MAX_PAYLOAD_SIZE_EXCEEDED';
+  message: string;
+}
+
 // This is sent when the server encounters an internal error
 // i.e. an invariant has been violated
 interface;
 
-type ProtocolError = UncaughtError | InvalidRequestError | CancelError;
+type ProtocolError =
+  | UncaughtError
+  | InvalidRequestError
+  | CancelError
+  | MaxPayloadSizeExceededError;
 ```
 
 `ProtocolError`s, just like service-level errors, are wrapped with a `Result`, which is further wrapped with `TransportMessage` and MUST have a `StreamCancelBit` flag. Please note that these are separate from user-defined errors, which should be treated just like any response message.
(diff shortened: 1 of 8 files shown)


cbrewster:

* main:
  * **PR #278**
    * **PR #279** 👈

router/client.ts
```diff
@@ -297,11 +299,42 @@ function handleProc(
...
+        // TODO: Is this the right error to send to the server?
+        sessionScopedSend(
+          cancelMessage(
+            streamId,
+            Err({
+              code: CANCEL_CODE,
+              message: 'cancelled by client',
+            }),
+          ),
+        );
```
cbrewster, line 334: Not sure if this is the right message, to send, maybe it should send the payload size exceeded error to the server too? Currently the server only ever expects `CANCEL_CODE` here

router/client.ts
```diff
@@ -297,11 +299,42 @@ function handleProc(
...
+
+        throw e;
```
cbrewster, line 336: Seems odd to have to throw here in addition to injecting the errors, but we do need a way for calls to `write` to error out if we hit this.

transport/sessionStateMachine/common.ts
```diff
@@ -141,6 +141,15 @@ export interface SessionOptions {
    * The codec to use for encoding/decoding messages over the wire
    */
   codec: Codec;
+  /**
+   * The maximum payload size that is allowed to be sent to the peer.
+   * This does not enforce max payload side when receiving messages,
```
cbrewster, line 146: Thoughts on enforcing this on received messages too? I think we get the most value out of this by just enforcing it on the sender side as long as we do that everything (server, client, python client)

PROTOCOL.md
```diff
@@ -672,3 +683,23 @@ This explicit ack serves three purposes:
...
+
+If the init message of a procedure exceeds the maximum payload size, the client must inject a `MAX_PAYLOAD_SIZE_EXCEEDED` error and fail the procedure. In this case no message is ever sent to the server and is handled entirely client-side.
+
+###### Non-init message
+
+If the client attempts to send a message after a procedure has been started which exceeds the maximum payload size. The client must throw an error inside the `write` call and inject a `MAX_PAYLOAD_SIZE_EXCEEDED` error in the procedure read stream. The client must also send a cancel request to the server so the server-side resources can be cleaned up.
+
+##### Server handling
+
+If the server attempts to send a message after a procedure has been started which exceeds the maximum payload size, the server must reply with a `MAX_PAYLOAD_SIZE_EXCEEDED` error and close down the procedure.
```
jackyzha0, line 705: is there a reason we need a new protocol error? could we just use CANCEL and have max payload size exceeded in the message? the semantics of MAX_PAYLOAD_SIZE_EXCEEDED and CANCEL are identical in how they should be handled i think

transport/sessionStateMachine/common.ts
jackyzha0: makes sense to me! we can probs do that follow up though, in our case we control all the clients so enforcing sender side is probably fine

router/client.ts
```diff
@@ -480,16 +513,33 @@ function handleProc(
...
-    streamId,
-    serviceName,
-    procedureName,
-    tracing: getPropagationContext(ctx),
-    payload: init,
-    controlFlags: procClosesWithInit
-      ? ControlFlags.StreamOpenBit | ControlFlags.StreamClosedBit
-      : ControlFlags.StreamOpenBit,
-  });
+  try {
```
jackyzha0, line 516: we can probably handle this uniformly across server/client in https://github.com/replit/river/blob/1007a553a24291c4d80cbcd420620003803eff4f/transport/transport.ts#L314

jackyzha0: try catch the send and if it errors, send cancel instead

PROTOCOL.md
cbrewster: Good point! My main goal was to make this error obvious when it happens, putting it in the message should be sufficient.

PR #283, opened 2025-01-31
Include river errors in server side spans

cbrewster:
## Why

This includes errors on the spans generated by the server so we have more error info.

I also changed the names of the spans to match the format that river-python is using.

In order to write a test for this, I had to make it so the tracer is created per-transport rather than having a global one.

## What changed

- De-global the tracer. The transport will construct and own a tracer instance (obtained from the global tracer provider).
- Add river errors to the server side spans
- Update the server/client spans to match the format we are using for river-python
    - This may make it harder to look at historical span data, let me know if you think this is a bad idea and I can revert it

## Versioning

### DIFF
--- a/router/client.ts
+++ b/router/client.ts
@@ -288,6 +288,7 @@ function handleProc(
   const procClosesWithInit = procType === 'rpc' || procType === 'subscription';
   const streamId = generateId();
   const { span, ctx } = createProcTelemetryInfo(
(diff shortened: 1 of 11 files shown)


cbrewster:

* main:
  * **PR #283** 👈

jackyzha0 (approved):

PR #288, opened 2025-02-14
rename session connect/disconnect -> created/closing/closed

jackyzha0:
## Why

nomenclature of 'connected'/'disconnected' is an artifact of when sessions were what connections are today, thats no longer accurate

this also allows us to disambiguate the 'about to close' state and 'already closed' state

these have two different cases:
1. about to close -> do any last minute cleanup and send things to the otherside before we totally yank the connection
2. already closed -> consumers can do things like manually open a new session if an old session is closed

## What changed

1. rename 'connect' -> 'created', 'disconnect' -> 'closing'
2. add a new 'closed' state which is emitted after the cleanup is done
3. fix a case where we could loop infinitely by mutating a map while iterating its values
4. add a test for the create a new session after previous session is gone case

## Versioning

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@replit/river",
   "description": "It's like tRPC but... with JSON Schema Support, duplex streaming and support for service multiplexing. Transport agnostic!",
-  "version": "1.0.0",
+  "version": "0.205.0",
   "type": "module",
   "exports": {
     ".": {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@replit/river",
   "description": "It's like tRPC but... with JSON Schema Support, duplex streaming and support for service multiplexing. Transport agnostic!",
-  "version": "0.204.0",
+  "version": "1.0.0",
   "type": "module",
   "exports": {
     ".": {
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@replit/river",
   "description": "It's like tRPC but... with JSON Schema Support, duplex streaming and support for service multiplexing. Transport agnostic!",
-  "version": "0.204.0",
+  "version": "0.205.0",
```
lhchavez, line 4: :one: wen

if there's no good answer, let's do it on Apr 1 (pls also blogpost)

testUtil/fixtures/matrix.ts
```diff
@@ -31,7 +31,9 @@ export const testMatrix = (selector?: Selector): Array<TestMatrixEntry> =>
     .map((transport) =>
       // If a selector is provided, filter transport + codecs to match the selector; otherwise, use all codecs.
       (selector
-        ? codecs.filter((codec) => selector[1] === codec.name)
+        ? codecs.filter((codec) => {
+            return selector[0] === transport.name && selector[1] === codec.name;
```
lhchavez, line 35: can you destructure `selector` into `[transportName, codecName]`? these offsets are spoopy.

lhchavez (approved):

package.json
jackyzha0: yes i'd like to! faris thinks that we need more docs but tbd on what we still need there

PR #290, opened 2025-03-20
better validation errors, dont include all errors

jackyzha0:
## Why

previously, if you accidentally sent the wrong payload (e.g. mismatched service versions), you would get `expected requestData or control payload` which is a very unclear message given that you did provide `requestData`, its just wrong

## What changed

- tweaked the error message to be more straightforward
- include extras in the invalid message back to the client

## Versioning

### DIFF
--- a/router/errors.ts
+++ b/router/errors.ts
@@ -1,5 +1,6 @@
 import {
   Kind,
+  Static,
   TLiteral,
   TNever,
   TObject,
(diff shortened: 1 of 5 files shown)


router/errors.ts
```diff
@@ -67,6 +89,12 @@ export const ReaderErrorSchema = Type.Union([
   Type.Object({
     code: Type.Literal(INVALID_REQUEST_CODE),
     message: Type.String(),
+    extras: Type.Optional(
```
masad-frost, line 92: Maybe this is spicy, but how do we feel about keeping this a record instead of an object? Just feels like extras shouldn't necessarily have meaning.

masad-frost (approved):

router/errors.ts
jackyzha0: just for this type if u use it directly, in the union its absorbed by the record type

blast-hardcheese (approved):

PR #296, opened 2025-03-25
fix sessionStatus type, remove no connection -> no connection transition on server, extra invariant checks

jackyzha0:
## Why

we get a very very low volume of message out of order (~40 messages across 15m connections) but these still suck

lets see if the message ordering problem is present before we send too

## What changed

add some extra invariant violation checks before message send

fix some small session status type alignment and removed an extra state where we would emit an extra session transition for the transparent reconnect case

## Versioning


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@replit/river",
   "description": "It's like tRPC but... with JSON Schema Support, duplex streaming and support for service multiplexing. Transport agnostic!",
-  "version": "0.205.1",
+  "version": "0.205.2",
```
lhchavez, line 4: this feels a bit minory: some args of callbacks change.

transport/sessionStateMachine/common.ts
```diff
@@ -224,15 +225,29 @@ export abstract class IdentifiedSession extends CommonSession {
    */
   seq: number;
 
+  /**
+   * Number of unique messages we've sent over the wire this session (excluding handshake)
+   */
+  messagesSent: number;
```
lhchavez, line 231: does this also include retransmitted messages?

lhchavez (approved):
seems good to me, but you might benefit from another, more in-depth reviewer.

package.json
jackyzha0: the interface change is backwards compatible (we passed the session field when the type didnt say it had a session field, now we declare a new id field which we do pass properly)

transport/server.ts
```diff
@@ -395,9 +395,8 @@ export abstract class ServerTransport<
           );
 
         oldSession = noConnectionSession;
+        this.updateSession(oldSession);
```
zhounathan, line 398: are there any implications for moving this update up into the if block?

PR #298, opened 2025-03-27
small type updates: unexport unwrapOrThrow, add regression test for ctx.cancel return

jackyzha0:
## Why

- people kept misusing unwrapOrThrow in production procedures

## What changed

- unexport unwrapOrThrow
- add a simple regression test for ctx.cancel in a union-ed case

## Versioning

### DIFF
--- a/router/index.ts
+++ b/router/index.ts
@@ -49,7 +49,7 @@ export type {
   ServiceContext,
   ProcedureHandlerContext,
 } from './context';
-export { Ok, Err, unwrapOrThrow } from './result';
+export { Ok, Err } from './result';
 export type {
   Result,
   ErrResult,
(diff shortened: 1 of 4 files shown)


router/index.ts
```diff
@@ -49,7 +49,7 @@ export type {
   ServiceContext,
   ProcedureHandlerContext,
 } from './context';
-export { Ok, Err, unwrapOrThrow } from './result';
+export { Ok, Err } from './result';
```
masad-frost, line 52: By misusing you mean they never attempt to handle expected errors/laziness? That's probably right, and I might be an offender.

masad-frost (approved):
I don't see a types update!

masad-frost:
oh i see what you mean

daweifeng-replit (approved):

PR #300, opened 2025-03-30
add more debug for message ordering violation

jackyzha0:
## Why

still getting `invariant violation: would have sent out of order msg`

## What changed

track last 10 sent messages along with their stack trace

this unfortunately degrades message throughput from ~46krps to ~16kps but should be more than enough for prod as is, we just need to find the bug and fix it and then remove this again

## Versioning

### DIFF
--- a/transport/sessionStateMachine/SessionConnected.ts
+++ b/transport/sessionStateMachine/SessionConnected.ts
@@ -84,7 +84,6 @@ export class SessionConnected<
         },
       });
 
-      this.closeConnection();
       throw new Error(msg);
     }
   }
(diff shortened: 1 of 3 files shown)


transport/sessionStateMachine/SessionConnected.ts
```diff
@@ -156,13 +183,6 @@ export class SessionConnected<
     });
   }
 
-  closeConnection() {
```
jackyzha0, line 159: this actually had the wrong semantics, it closed the connection but removed the close listener so that the session would never know the connection was closed

blast-hardcheese: do we need to clean these up elsewhere?

blast-hardcheese (approved):

masad-frost (approved):
Muh RPS

Hopefully that's all we need. Strange buggggg

PR #302, opened 2025-04-01
make client-side heartbeat timeout instead of interval based

jackyzha0:
## Why

- https://developer.chrome.com/blog/timer-throttling-in-chrome-88 causes intervals to be heavily throttled, prefer to use a single timeout

## What changed

- refactor to use a single timeout instead of heartbeatsUntilDead * intervals
- fix some mock transport harness logic
- remove stack tracking code, it didnt tell us much 😭 

## Versioning

### DIFF
--- a/transport/sessionStateMachine/common.ts
+++ b/transport/sessionStateMachine/common.ts
@@ -281,10 +281,6 @@ export abstract class IdentifiedSession extends CommonSession {
   constructMsg<Payload>(
     partialMsg: PartialTransportMessage<Payload>,
   ): TransportMessage<Payload> {
-    if (this._isConsumed) {
-      throw new Error(ERR_CONSUMED);
-    }
-
     const msg = {
       ...partialMsg,
       id: generateId(),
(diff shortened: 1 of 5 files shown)


transport/sessionStateMachine/common.ts
```diff
@@ -281,10 +281,6 @@ export abstract class IdentifiedSession extends CommonSession {
   constructMsg<Payload>(
     partialMsg: PartialTransportMessage<Payload>,
   ): TransportMessage<Payload> {
-    if (this._isConsumed) {
-      throw new Error(ERR_CONSUMED);
-    }
-
```
zhounathan, line 287: why don't we error on this anymore?

zhounathan (approved):

transport/sessionStateMachine/common.ts
jackyzha0: i added this as a debug for sanity checking two patches ago but we never hit it so its a redundant check

masad-frost (approved):

PR #303, opened 2025-04-01
account for RangeError: Array buffer allocation failed

jackyzha0:
## Why

```ts
function replicateArrayBufferError() {
  try {
    const size = 2 * 1024 * 1024 * 1024;
    const buf = new ArrayBuffer(size);
	conn.send(buf);
  } catch (e) {
    console.error(e);
  }
}

replicateArrayBufferError();
console.log('you can try catch this!!!')
```

tldr; a failed allocation can be try-caught by a top-level handler (e.g. in react) and river can pretend nothing went wrong and construct another message

by the time the next message comes around, its possible theres enough memory to allocate again and whoops!!! we skipped sending a message

some more context on this error: https://groups.google.com/a/chromium.org/g/blink-dev/c/pZ9kld0LehA/m/Ho-SCn8tBAAJ

## What changed

- assume codecs are fallible: wrap codec in an adapter (`CodecMessageAdapter`) to try catch toBuffer and fromBuffer operations
- delete the session if codec fails, this should gc the message buffer for that session which should help memory pressure
- we also expose `ProtocolError.MessageSendFailure` so the consumer can decide to shut down the world in this case too
- modify ws connection to not throw when sending on `CLOSING` or `CLOSED` states (safe to discard as we have the sendbuffer)
- actually read the result of .send()
- added test to prevent regression
- delete unused `messageFraming` code (we axed UDS a while ago)

## Versioning


transport/connection.ts
```diff
@@ -116,7 +116,7 @@ export abstract class Connection {
    * @param msg The message to send.
    * @returns true if the message was sent, false otherwise.
    */
-  abstract send(msg: Uint8Array): boolean;
+  abstract send(msg: Uint8Array): boolean | void;
```
masad-frost, line 119: no need for | void

transport/impls/ws/connection.ts
```diff
@@ -52,13 +52,13 @@ export class WebSocketConnection extends Connection {
   }
 
   send(payload: Uint8Array) {
-    if (this.ws.readyState !== this.ws.OPEN) {
+    try {
+      this.ws.send(payload);
+
+      return true;
+    } catch {
       return false;
```
masad-frost, line 60: I'm not sure how i feel about absolutely swallowing everything here. Should we log the error?

transport/sessionStateMachine/common.ts
```diff
@@ -156,42 +157,18 @@ export abstract class CommonSession extends StateMachineState {
...
-  constructor({ from, options, log, tracer }: CommonSessionProps) {
+  constructor({ from, options, log, tracer, codec }: CommonSessionProps) {
     super();
     this.from = from;
     this.options = options;
     this.log = log;
     this.tracer = tracer;
-  }
-
-  parseMsg(msg: Uint8Array): OpaqueTransportMessage | null {
```
masad-frost, line 171: RIP

codec/adapter.ts
```diff
@@ -0,0 +1,48 @@
+import { Value } from '@sinclair/typebox/value';
+import {
+  OpaqueTransportMessage,
+  OpaqueTransportMessageSchema,
+} from '../transport';
+import { Codec } from './types';
+import { DeserializeResult, SerializeResult } from '../transport/results';
+import { coerceErrorString } from '../transport/stringifyError';
+
+export class CodecMessageAdapter {
```
masad-frost, line 10: we should leave doc strings on all of our classes!

masad-frost: curious why an adapter instead of making the implementation conform to this

transport/client.ts
```diff
@@ -292,13 +292,46 @@ export abstract class ClientTransport<
...
+
+          this.protocolError({
+            type: ProtocolError.MessageSendFailure,
+            message: reason,
+          });
+          this.deleteSession(connectedSession, { unhealthy: true });
+        },
+      });
+
+    const res = connectedSession.sendBufferedMessages();
```
masad-frost, line 320: good move, I'd go as far as removing the `startMissingHeartbeatTimeout` from the constructor and make it the responsibility of the object creator, too. Side-effect free constructors tend to make things more predictable and testable. You can put both methods behind the same function (i.e. `init`)

masad-frost (approved):
FIVE BIG BOOMS

:boom: 
:boom: 
:boom: 
:boom: 
:boom: 

tiny comments

transport/impls/ws/connection.ts
jackyzha0: this only ever fails for one reason, sending on CONNECTING state, imo its fine? connection doesnt have direct access to a logger and i think threading it through just for this doesnt super make sense

we can alternatively trigger this.ws.onerror and close

transport/server.ts
```diff
@@ -456,9 +470,24 @@ export abstract class ServerTransport<
...
+        type: ProtocolError.MessageSendFailure,
+        message: res.reason,
+      });
+      this.deletePendingSession(session);
+
+      return;
+    }
 
     // transition
+    this.pendingSessions.delete(session);
```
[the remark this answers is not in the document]
jackyzha0: this is the point where we 'consume' the pending session and turn it into a connected session

now that there is a fail point, we want to clear the old pending session earlier rather than later

codec/adapter.ts
jackyzha0: codec is user substitutable and we cant expect every consumer to remember to try/catch toBuffer fromBuffer, so we do it at this layer

PR #304, opened 2025-04-02
dont need multiple listeners on connection

jackyzha0:
## Why

the connection class used to support a set of listeners for each type, YAGNI! and can conceal potential failures where we are dispatching events to multiple sessions

## What changed

make listeners `cb | undefined` instead of `Set<cb>`

## Versioning


zhounathan (approved):
everything makes sense to me, since it's a transport-level change I assume compatibility is irrelevant? older connections still use the previous transport cb list and new ones will just use callbacks

PR #308, opened 2025-04-08
Add read-only middleware support

daweifeng-replit:
## Why

This PR adds support of read-only middleware support. Before a procedure is called, river will go through a chain of middleware synchronously and the middlewares are read-only so they shouldn't modify the context.

## What changed

- Add `middlewares` in server options
- Chain and run middlewares before procedure calls
- Add middleware tests

## Versioning

### DIFF
--- a/__tests__/middleware.test.ts
+++ b/__tests__/middleware.test.ts
@@ -194,21 +194,13 @@ describe('middleware test', () => {
 
   test('apply multiple middlewares in order', async () => {
     const services = { test: TestServiceSchema };
-    // counter for checking the call order
-    let callOrder = 0;
     const middleware1 = vi.fn(({ next }: MiddlewareParam) => {
-      callOrder++;
-      expect(callOrder).toBe(1);
       next();
     });
     const middleware2 = vi.fn(({ next }: MiddlewareParam) => {
-      callOrder++;
-      expect(callOrder).toBe(2);
       next();
     });
     const middleware3 = vi.fn(({ next }: MiddlewareParam) => {
-      callOrder++;
-      expect(callOrder).toBe(3);
       next();
     });
     createServer(mockTransportNetwork.getServerTransport(), services, {
@@ -220,6 +212,13 @@ describe('middleware test', () => {
     );
 
     const result = await client.test.add.rpc({ n: 3 });
+
+    expect(middleware1.mock.invocationCallOrder[0]).toBeLessThan(
+      middleware2.mock.invocationCallOrder[0],
+    );
+    expect(middleware2.mock.invocationCallOrder[0]).toBeLessThan(
+      middleware3.mock.invocationCallOrder[0],
+    );
     expect(result).toStrictEqual({ ok: true, payload: { result: 3 } });
   });
 });
--- a/__tests__/middleware.test.ts
+++ b/__tests__/middleware.test.ts
@@ -8,6 +8,7 @@ import {
 } from '../testUtil/fixtures/services';
 import { createClient, createServer } from '../router';
 import { createMockTransportNetwork } from '../testUtil/fixtures/mockTransport';
+import { MiddlewareParam } from '../router/server';
 
 describe('middleware test', () => {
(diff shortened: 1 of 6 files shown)


router/server.ts
```diff
@@ -586,36 +587,37 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+            middleware({
+              ctx: middlewareContext,
+              reqInit: initPayload,
+              next,
+            });
+          };
+        },
+        () => {
+          // All middleware complete
+        },
```
masad-frost, line 608: if the last "next" is an empty function and not the middleware then `AsyncLocalStorage.run` would not work as expected

masad-frost: i.e.

  storage.run('hello', () => {});

  const value = storage.getStore();
  console.log(value); // logs undefined

daweifeng-replit: That's right. I am updating it so it can call handlers in the last step

router/server.ts
```diff
@@ -681,9 +659,29 @@ class RiverServer<Services extends AnyServiceSchemaMap>
...
+            });
+          };
+        },
+        () => {
+          void handlerToRun();
+        },
+      )(); // Immediately invoke the generated middleware chain function
+    };
+
+    // Start the middleware chain, which will eventually call runProcedureHandler
```
masad-frost, line 683: applyMiddlewares feels unnecessary, let's remove it and call the chain immediately?

__tests__/middleware.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    });
+    const client = createClient<typeof services>(
+      mockTransportNetwork.getClientTransport('client'),
+      'SERVER',
+    );
+
+    const result = await client.test.add.rpc({ n: 3 });
+    expect(result).toStrictEqual({ ok: true, payload: { result: 3 } });
+  });
+});
```
masad-frost, line 225: Can we add another test with calling next within AsyncLocalStorage.run please?

__tests__/middleware.test.ts
```diff
@@ -0,0 +1,225 @@
...
+        }),
+        reqInit: {},
+      }),
+    );
+  });
+
+  test('apply multiple middlewares in order', async () => {
+    const services = { test: TestServiceSchema };
+    // counter for checking the call order
+    let callOrder = 0;
```
masad-frost, line 198: FYI you can also do this https://github.com/replit/river/blob/2f005b5113a63c5829383e884691af87b4b4cce9/transport/events.test.ts#L41-L43

masad-frost: Actually you should definitely do that.

Putting expect calls within a mock function that you're passing to a library may never run and your assertions are never checked.

router/server.ts
```diff
@@ -1024,15 +1022,17 @@ interface MiddlewareContext
...
 /**
  * Middleware is a function that can inspect requests as they are received.
  * For now modification of the request is not supported behavior.
  */
-type Middleware = (param: {
-  readonly ctx: MiddlewareContext;
-  readonly reqInit: Static<PayloadType>;
-  next: () => void;
-}) => void;
+export type Middleware = (param: MiddlewareParam) => void;
```
masad-frost, line 1035: Maybe we should export this via router/index

__tests__/middleware.test.ts
daweifeng-replit: I added a test that shows how AsyncLocalStorage can persist the context when there is a `unhandledRejection`.

daweifeng-replit: I am also trying to write a test case when an error is thrown from `setTimeout`. However, AsyncLocalStorage is empty in `uncaughtException`. I am wondering if `setTimeout` is a special case that AsyncLocalStorage does not support.

daweifeng-replit: Seems like `asyncLocalStorage.enterWith` would work in the `setTimeout` test

masad-frost (approved):

PR #312, opened 2025-04-16
fix(json codec): encode the empty buffer properly

jackyzha0:
## Why

`{ $t: "" }` means that `$t` is false-y so we dont treat it as the buffer type

## What changed

make it a strict undefined check

## Versioning


daweifeng-replit (approved):
LGTM

PR #314, opened 2025-05-01
Update typebox to support ~0.34.0

daweifeng-replit:
## Why

The latest TypeBox introduces Module and Ref. We should support that.

## What changed

- Change `@sinclair/typebox` peerDependency to `~0.32.8 || ~0.34.0`

## Versioning

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@replit/river",
   "description": "It's like tRPC but... with JSON Schema Support, duplex streaming and support for service multiplexing. Transport agnostic!",
-  "version": "0.208.1",
+  "version": "0.208.2",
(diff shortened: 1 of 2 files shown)


jackyzha0 (approved):
i think we can just bump directly

package.json
```diff
@@ -54,7 +54,7 @@
   },
   "peerDependencies": {
     "@opentelemetry/api": "^1.7.0",
-    "@sinclair/typebox": "~0.32.8"
+    "@sinclair/typebox": "~0.32.8 || ~0.34.0"
```
jackyzha0, line 57: oh i meant make this just `~0.34.0` + we upgrade web

PR #318, opened 2025-05-07
[chore] Adding release-drafter config

blast-hardcheese:
## Why

Release drafter workflow would be nice here

## Versioning


jackyzha0 (approved):
weee

PR #322, opened 2025-06-09
Refactor ServiceContext as generic

daweifeng-replit:
## Why

When we are creating a server using river, we have to declare the `ServiceContext` interface globally so TypeScript can infer the type for `ctx` within a procedure. 

```js
 declare module '@replit/river' {
   interface ServiceContext {
       db: Database;
   }
 }
```

However, it will be an issue in a mono repo setup where we want to create multiple servers with different `ServiceContext`. 

## What changed

Use builder pattern for `ServiceSchema` so we can create a `ServiceSchema` with Context like 

`const ServiceSchema = createServiceSchema<Context>();`

Example: 

```js

const extendedContext = {
  testctx: Math.random().toString(),
};

const ServiceSchema = createServiceSchema<typeof extendedContext>();

const services = {
  testservice: ServiceSchema.define({
    testrpc: Procedure.rpc({
      requestInit: Type.Object({}),
      responseData: Type.String(),
      handler: async ({ ctx }) => {
        return Ok(ctx.testctx);
      },
    }),
  }),
};
```

When we create a server, it will perform type checking to make sure that all the services are expecting the same `ServiceContext` as the type of `extendedContext` in the config.

```js
createServer(serverTransport, services, {
  extendedContext,
});
```

There is no type-checking for `Context` in createClient(). We are using `any` to skip it behind the scene because it is a implementation detail on the server side.

```js
const client = createClient<typeof services>(
  clientTransport,
  serverTransport.clientId,
); 
```

## Versioning

### DIFF
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -43,10 +43,10 @@ describe('should handle incompatabilities', async () => {
       testctx: Math.random().toString(),
     };
 
-    const serviceSchema = createServiceSchema<ExtendedContext>();
+    const ServiceSchema = createServiceSchema<ExtendedContext>();
 
     const services = {
-      testservice: serviceSchema.define({
+      testservice: ServiceSchema.define({
         testrpc: Procedure.rpc({
           requestInit: Type.Object({}),
           responseData: Type.String(),
@@ -82,9 +82,9 @@ describe('should handle incompatabilities', async () => {
       testctx: string;
     }
 
-    const serviceSchema = createServiceSchema<ExtendedContext>();
+    const ServiceSchema = createServiceSchema<ExtendedContext>();
 
-    const TestServiceScaffold = serviceSchema.scaffold({
+    const TestServiceScaffold = ServiceSchema.scaffold({
       initializeState: (ctx) => ({
         fromctx: ctx.testctx,
       }),
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -40,23 +40,30 @@ describe('should handle incompatabilities', async () => {
     // setup
     const clientTransport = getClientTransport('client');
     const serverTransport = getServerTransport();
+
+    interface ExtendedContext {
+      testctx: string;
+    }
+    const extendedContext: ExtendedContext = {
+      testctx: Math.random().toString(),
+    };
+
     const services = {
-      testservice: ServiceSchema.define({
+      testservice: ServiceSchema.defineWithContext<ExtendedContext>()({
         testrpc: Procedure.rpc({
           requestInit: Type.Object({}),
           responseData: Type.String(),
           handler: async ({ ctx }) => {
-            return Ok((ctx as unknown as typeof extendedContext).testctx);
+            return Ok(ctx.testctx);
           },
         }),
       }),
     };
 
-    const extendedContext = { testctx: Math.random().toString() };
     createServer(serverTransport, services, {
       extendedContext,
     });
-    const client = createClient<typeof services>(
+    const client = createClient<typeof services, ExtendedContext>(
       clientTransport,
       serverTransport.clientId,
     );
--- a/__tests__/context.test.ts
+++ b/__tests__/context.test.ts
@@ -5,14 +5,9 @@ import {
 } from '../testUtil/fixtures/cleanup';
 import { testMatrix } from '../testUtil/fixtures/matrix';
 import { TestSetupHelpers } from '../testUtil/fixtures/transports';
-import {
-  Ok,
-  Procedure,
-  ServiceSchema,
-  createClient,
-  createServer,
-} from '../router';
+import { Ok, Procedure, createClient, createServer } from '../router';
 import { Type } from '@sinclair/typebox';
+import { createServiceSchema } from '../router/services';
 
 describe('should handle incompatabilities', async () => {
   const { addPostTestCleanup, postTestCleanup } = createPostTestCleanups();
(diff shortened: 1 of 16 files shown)


Monkatraz:
Ah yeah Jacky's idea is better, we basically want a server builder scaffold

__tests__/context.test.ts
```diff
@@ -40,19 +35,28 @@ describe('should handle incompatabilities', async () => {
...
     const serverTransport = getServerTransport();
+
+    interface ExtendedContext {
+      testctx: string;
+    }
+    const extendedContext: ExtendedContext = {
+      testctx: Math.random().toString(),
+    };
+
+    const ServiceSchema = createServiceSchema<ExtendedContext>();
```
jackyzha0, line 46:
```suggestion
    const extendedContext = {
      testctx: Math.random().toString(),
    };

    const ServiceSchema = createServiceSchema(extendedContext);
```

or alternatively

```suggestion
    const extendedContext = {
      testctx: Math.random().toString(),
    };

    const ServiceSchema = createServiceSchema<typeof extendedContext>();
```

though i prefer the first

__tests__/typescript-stress.test.ts
```diff
@@ -61,7 +64,7 @@ const fnBody = Procedure.rpc<
 // typescript is limited to max 50 constraints
 // see: https://github.com/microsoft/TypeScript/issues/33541
 // we should be able to support more than that due to how we make services
-const StupidlyLargeServiceSchema = ServiceSchema.define({
+const StupidlyLargeServiceSchema = createServiceSchema().define({
```
jackyzha0, line 67: we should test context in this large case too

router/services.ts
```diff
@@ -30,17 +30,22 @@ export interface Service<
...
+>;
 
 /**
  * A dictionary of {@link ServiceSchema}s, where the key is the service name.
  */
-export type AnyServiceSchemaMap = Record<string, AnyServiceSchema>;
+export type AnyServiceSchemaMap<Context extends object = object> = Record<
+  string,
+  AnyServiceSchema<Context>
+>;
```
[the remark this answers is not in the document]
daweifeng-replit: The reason we need `Context` for  `AnyServiceSchemaMap` is the same reason as the conversation above. TypeScript would complain when we try to assign a non-empty `Context` to `{}`

__tests__/context.test.ts
daweifeng-replit: I have changed it to `const ServiceSchema = createServiceSchema(extendedContext);`

router/services.ts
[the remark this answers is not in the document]
daweifeng-replit: @Monkatraz Is there a way to define a object-like TypeScript type?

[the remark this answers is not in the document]
Monkatraz: `object` typically works

Monkatraz:
Do we have to pass the extended context as a parameter? That goes against how e.g. Chort defines its services and will require a significant refactor

__tests__/typescript-stress.test.ts
daweifeng-replit: Test added. Procedures can access context in the test

daweifeng-replit:
> Do we have to pass the extended context as a parameter? That goes against how e.g. Chort defines its services and will require a significant refactor
> 
> I think it goes against the spirit of River's service schema if you have to instantiate a bunch of potential side effects just to construct the schema

There is another option which is

```js
const extendedContext = {
    testctx: Math.random().toString(),
};
const ServiceSchema = createServiceSchema<typeof extendedContext>();
```

Would this work for Chort?

jackyzha0:
yeah now that i think of it just having the type and not constructing the whole object is probably better, lets go with the second approach

jackyzha0:
id like to see if we can spend a bit more time avoiding the `any` before we merge, it scares me and tells us the types are flowing wrong somewhere

daweifeng-replit:
> id like to see if we can spend a bit more time avoiding the `any` before we merge, it scares me and tells us the types are flowing wrong somewhere

Looking into the code. We might not be able to do it without `any`. Here is an example 
```js
type Handler<T> = (ctx: T) => void;

const specificHandler: Handler<{ db: string }> = (ctx) => {
  console.log(ctx.db); // Expects 'db' to exist
};

type AnyHandlerMap<Context extends object = object> = Record<string, Handler<Context>>

const services = {
  specificHandler
}

const generalServices: AnyHandlerMap<object> = services
```
TypeScript would complain about

```
Type '{ specificHandler: Handler<{ db: string; }>; }' is not assignable to type 'AnyHandlerMap<object>'.
  Property 'specificHandler' is incompatible with index signature.
    Type 'Handler<{ db: string; }>' is not assignable to type 'Handler<object>'.
      Property 'db' is missing in type '{}' but required in type '{ db: string; }'. (typescript)
```

Because AnyHandlerMap is not actual accepting "Any" procedures unless we set it to `AnyHandlerMap<any>`

daweifeng-replit:
As we discussed offline, we are going to keep `any` in the `Context` param of `createClient()` and `Client` generic. One potential idea to fix that is to create a new `ClientService` type that does not require `Context` so we can map `ServiceSchema` to `ClientService`. We can try it in the future.

daweifeng-replit:
* **#322** <a href="https://app.graphite.dev/github/pr/replit/river/322?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/replit/river/322?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #327, opened 2025-06-16
Update Github Workflow

daweifeng-replit:
## Why

The previous `release drafter` and `Build and upload package` workflow did not work! In the previous setup, when we publish a draft release created by `release drafter`, it does not update the version in `package.json` so the `package.json` file in the released source code is outdated. 

## What changed

- We still use release drafter like before. But instead of publishing the draft release directly, we should Go to Actions -> workflows/prepare-release.yml and run the workflow.
- It will automatically create a git tag from the draft release 
- The new git tag will trigger the `publish.yml` workflow, which will build and publish the package to npm

## Versioning


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,51 @@
...
+        id: release_drafter
+        uses: release-drafter/release-drafter@v6
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # Use the output from the previous step to version the package
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
jackyzha0, line 30: i think this is deprecated?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,36 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        # This checks out the code at the specific tag that triggered the workflow
+        uses: actions/checkout@v4
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 18
```
jackyzha0, line 20: same here

jackyzha0 (approved):

PR #330, opened 2025-06-17
Pass ParsedMetadata to SeviceSchema.scaffold()

daweifeng-replit:
## Why

I should have pass `ParsedMetadata` to `SeviceSchema.scaffold()`. Otherwise, the `ServiceScaffold` always sets `ParsedMetadata` as `object` by default

## What changed

- pass `ParsedMetadata` to `SeviceSchema.scaffold()`
- update tests 

## Versioning

### DIFF
--- a/__tests__/e2e.test.ts
+++ b/__tests__/e2e.test.ts
@@ -954,8 +954,14 @@ describe.each(testMatrix())(
         await cleanupTransports([clientTransport, serverTransport]);
       });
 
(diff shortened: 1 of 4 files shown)


masad-frost (commented):
That's precisely why I think we should be passing around <HandlerContext> IMO, we wouldn't miss these things

daweifeng-replit:
> Unblocking for now, but if you have a few minutes to refactor it would be nice

Agree! I will refactor it in a separate PR.

PR #332, opened 2025-07-15
export ReadableBrokenError, RawReadable, ReadableResult

jackyzha0:
## Why

- nice to have these for tests instead of importing `it-pushable`
- missed a type export

## What changed

- re-export above types

## Versioning


masad-frost (approved):
Some of the methods have @internal on their comments, we should probably remove that and do a general pass on the comments.

Issue #333, opened 2025-07-15
Allow "throwing" in server streams


masad-frost:
`UNCAUGHT_ERROR` is built into the protocol, it means the server encountered an internal error in processing the request.

We've noticed that people use this correctly when using unary rpcs but not all the time in streaming rpcs. The reason being is that if you're outside the main handler function (i.e. in a callback or background task), you have no way of triggering an internal error, throwing will not be caught by the handler.

This leads people to define error types that are basically unhandlable by the client and are effectively custom internal errors. This leads to bad error hygiene on the client side.

We should have a way to basically end the request with an uncaught, something like ctx.uncaught(new Error("something went wrong")), which would end the request following similar semantics as throwing within the handler.

PR #335, opened 2025-07-15
Add `ctx.uncaught` for emitting uncaught errors, particularly in streams

Monkatraz:
## Why

Fixes https://github.com/replit/river/issues/333

In streaming procedures, you often need to push an uncaught error into the stream _outside_ of the scope of the handler. This is tricky since all you're left with doing is manually returning an uncaught error rather than using any of the machinery River has built in.

## What changed

Added `ctx.uncaught` which works a lot like `ctx.cancel` except that it uses `UNCAUGHT_ERROR` rather than the cancel code. It uses the exact same mechanism as the normal procedure error handler and yields you an error result that you can return if needed. It accepts anything as an argument so that you can use it ergonomically in a try catch.

## Versioning

### DIFF
--- a/router/context.ts
+++ b/router/context.ts
@@ -2,7 +2,7 @@ import { Span } from '@opentelemetry/api';
 import { TransportClientId } from '../transport/message';
 import { SessionId } from '../transport/sessionStateMachine/common';
 import { ErrResult } from './result';
-import { CancelErrorSchema } from './errors';
+import { CancelErrorSchema, UncaughtErrorSchema } from './errors';
 import { Static } from '@sinclair/typebox';
 
 /**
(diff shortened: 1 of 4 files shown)


router/context.ts
```diff
@@ -40,6 +40,15 @@ export type ProcedureHandlerContext<State, Context, ParsedMetadata> =
      * the river documentation to understand the difference between the two concepts.
      */
     cancel: (message?: string) => ErrResult<Static<typeof CancelErrorSchema>>;
+    /**
+     * This emits an uncaught error in the same way that throwing an error in a handler
+     * would. You should minimize the amount of work you do after calling this function
```
daweifeng-replit, line 45: Maybe we can create an eslint rule for this

PR #339, opened 2025-09-02
add isFatalConnectionError (closes #282)

### LINKED ISSUE #282: add ability for `createNewOutgoingConnection` to explicitly stop retrying in certain failure modes
e.g. if a repl is taken down, we should stop retrying the connection to it

jackyzha0:
## Why

there are cases we dont want to reconnect!

## What changed

add a client option to determine what errors are 'fatal' and should not be considered retriable

## Versioning


masad-frost (approved):
sick

PR #359, opened 2026-02-23
enable codex review

jackyzha0:
## Why

codex helps catch bugs for us! this is an internal/private gh action but should be safe to run even though this is a public repo

low-ish risk as actions can only be triggered by replit employees

## What changed

add codex action

## Versioning


darshkpatel (commented):
Hmm, since this is a public repo we should only do this on maintainers PRs automatically and have a way for us to manually trigger it for other PRs to  prevent abuse

PR #379, opened 2026-05-31
Check for malformed ctx.from ids in clients

Monkatraz:
## Why

A broken client could send messages as the wrong `from` and cause things to break.

## What changed

- Made it so that sessions drop clients who send messages with the wrong from.
- Made it so that during `validate`, a server can check the transport identity and ensure binding of the ID is valid.

## Versioning

### DIFF
--- a/transport/sessionStateMachine/SessionConnected.ts
+++ b/transport/sessionStateMachine/SessionConnected.ts
@@ -192,6 +192,15 @@ export class SessionConnected<
 
     const parsedMsg = parsedMsgRes.value;
 
+    // messages must originate from this session's peer
+    if (parsedMsg.from !== this.to) {
+      this.listeners.onInvalidMessage(
+        `received message with 'from' (${parsedMsg.from}) that does not match the session peer (${this.to})`,
+      );
+
+      return;
+    }
+
     // check message ordering here
     if (parsedMsg.seq !== this.ack) {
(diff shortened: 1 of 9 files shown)


Monkatraz:
* **#379** <a href="https://app.graphite.com/github/pr/replit/river/379?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/replit/river/379?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

wernst (approved):

transport/sessionStateMachine/SessionConnected.ts
```diff
@@ -192,6 +192,15 @@ export class SessionConnected<
 
     const parsedMsg = parsedMsgRes.value;
 
+    // messages must originate from this session's peer
+    if (parsedMsg.from !== this.to) {
```
lhchavez, line 196: why do we even allow connections to specify the to/from? i think this is a vestigial feature that we never ended up using.

lhchavez (commented):
lgtm, but we should try to remove those fields

transport/sessionStateMachine/SessionConnected.ts
Monkatraz: I believe you're correct

PR #401, opened 2026-08-20
Explore adding structured handshake rejections

Mayank808:
## Why

Custom handshake handlers can only return a River failure code. Applications must put specific failure states in error strings and parse those strings for retry decisions.

## What changed

Custom handshake handlers can now return `rejectHandshake(extras)`, where `extras` is an opaque, application-defined value. River sends it on the optional `extras` field of failed handshake responses and exposes it in `handshake_failed` protocol error events and logs, without interpreting it. Applications define and validate their own schema for `extras`, so adding a new application error type does not require a River change. Existing handlers that return a River failure code continue to work without changes.

Tests cover initial handshakes, re-handshakes, and compatibility with old clients and servers. The protocol and handshake documentation describe the new field.

## Versioning

### DIFF
--- a/PROTOCOL.md
+++ b/PROTOCOL.md
@@ -258,13 +258,9 @@ interface ControlHandshakeResponse {
           // fatal, returned by the custom handshake handler
           | 'REJECTED_BY_CUSTOM_HANDLER'
           | 'REJECTED_UNSUPPORTED_CLIENT';
-        // Application-defined rejection details. Older peers ignore this
-        // optional field.
-        details?: {
-          code: string;
-          message: string;
-          extras?: unknown;
-        };
+        // Application-defined rejection data. River populates this only for
+        // custom-handler rejections. Older peers ignore this optional field.
+        extras?: unknown;
       };
 }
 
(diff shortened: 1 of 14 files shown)


PROTOCOL.md
```diff
@@ -258,6 +258,13 @@ interface ControlHandshakeResponse {
           // fatal, returned by the custom handshake handler
           | 'REJECTED_BY_CUSTOM_HANDLER'
           | 'REJECTED_UNSUPPORTED_CLIENT';
+        // Application-defined rejection details. Older peers ignore this
+        // optional field.
+        details?: {
```
wernst: What do you think about making this just `extras: unknown`? I think at this point we're at the application boundary and we dont need to be prescriptive about shape. This is up to the app dev.

wernst (approved):

wernst (approved):

PROTOCOL.md
Mayank808: I think that also works but might lead to a lot of application specific structures and duplication that we would have to keep track of. I think having some structure enforced in a details package could lead to bit more conformity in the codebase later on especially when this is only being used for error cases. But let me know this is a easy change and more about style preferences of the team

wernst: > application specific structures and duplication that we would have to keep track of

Can you expand on this?

PROTOCOL.md
```diff
@@ -258,6 +258,13 @@ interface ControlHandshakeResponse {
...
+          code: string;
```
wernst: As I understand it's a bit of an anti pattern in river to have `code: string`. It tends to prefer specified union types (see `code` in the layer above)

PROTOCOL.md
wernst: Does this details structure fit with every code or just with `REJECTED_BY_CUSTOM_HANDLER`? If not the type is currently a bit leaky (in that it specifies that details belongs with all codes - and maybe thats intented)

darshkpatel:
who consumes this and what do they do with it? construct already refetches tokens per attempt, and terminal stuff like repl gone can be explained by the api. wondering what case actually needs this on the wire

PR #408, opened 2026-09-05
Export nameable procedure definition types

replitom:
## Why

River's procedure constructors return a branded type whose public inferred declarations can expose the internal `__BRAND_DO_NOT_USE` property. Downstream packages with declaration emit enabled therefore cannot give exported procedure definitions and instantiated service maps stable public names without manually widening them.

## What changed

- Add public `ProcedureDefinition` and exact `ProcedureDefinitionMap` aliases while preserving the existing brand representation and constructor assignability.
- Export the existing schema-map and instantiated service-map types from the package root.
- Add a consumer declaration-emit regression test covering constructor assignability, exact procedure keys, payload preservation, public service-map naming, and absence of the internal brand name in the consumer declaration.

Downstream use depends on publishing a new `@replit/river` release and then bumping the consuming package's River version. This PR does not patch any consumer's `node_modules`.

## Versioning

~ written by Replit ჻

### DIFF
--- a/__tests__/fixtures/public-declaration-consumer.ts
+++ b/__tests__/fixtures/public-declaration-consumer.ts
@@ -1,0 +1,62 @@
+import { type Static, type TNever, Type } from 'typebox';
+import {
+  createServiceSchema,
+  type InstantiatedServiceSchemaMap,
+  Procedure,
+  type ProcedureDefinition,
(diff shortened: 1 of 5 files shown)


replitom:
Closing after measuring the result: this did not produce a large enough improvement in the developer experience to keep.

PR #412, opened 2026-09-14
Add support for returning raw protobufs directly

Armster15:
Lets protobuf handlers return already-encoded response bytes without another protobuf encode. Existing typed handlers keep working unchanged.

## Why

Some handlers already have encoded responses. This keeps the opt-in limited to raw protobuf bytes instead of adding a general serialization API.

## What changed

```ts
ProtoService.define(Greeter, {
  sayHello: {
    raw: 'output',
    handler: () => Ok(encodedResponse),
  },
});
```

- `raw: 'output'` keeps requests typed and sends successful response bytes unchanged.
- `raw: 'both'` also gives the handler an owned copy of the request bytes.
- Both modes support all four method kinds. Stream handlers write `Ok(bytes)` to `resWritable`.
- Middleware still receives protobuf messages. Clients, wire format, typed errors, and stream cleanup stay unchanged.

Responses must contain valid bytes for the declared protobuf message. Raw mode lives on each registration, without a cache or serde wrapper.
The shared `Codec` interface and existing typed registration helpers are unchanged.

## Validation

794 tests passed, with one existing skip. `npm run check` and `npm run build` passed.
The feature suite has 9 focused checks for types, byte preservation, input ownership, middleware, errors, and cleanup.
Mutation checks prove that the reduced suite detects re-encoding, missing input copies, lost registration modes, and broken error handling.

## Versioning

No version bump or publication is included. After adoption, callers must remove raw registrations before downgrading River.

## Stack

Merge bottom-to-top.

1. #411: test-tooling prerequisite (1/3).
2. #412: raw protobuf response support (2/3), this PR.
3. #414: bump River to 0.223.0 (3/3).

~ written by Zerg :space_invader: ([aggressive-thor-eeb8](https://zerg.zergrush.dev/chat?id=aggressive-thor-eeb8), [wp-a4078f5f](https://zerg.zergrush.dev/chat?id=wp-a4078f5f))


__tests__/protobuf-raw.test.ts
```diff
@@ -0,0 +1,627 @@
...
+  Context extends object,
+  State extends object,
+> {
+  readonly descriptor: Service;
+  readonly methods: ReadonlyMap<string, MainRegisteredMethod>;
+  readonly initializeStateFn:
+    | ((ctx: Context) => MainMaybeDisposable<State>)
+    | undefined;
+  instantiate(ctx: Context): MainInstantiatedProtoService;
+}
```
jackyzha0: sus that we need to redefine these types

README.md
```diff
@@ -939,6 +939,28 @@ const transport = new WebSocketClientTransport(
...
+
+Raw handlers keep the typed handler's signature, but receive `Uint8Array` requests and use `Ok(Uint8Array)` for successful responses.
+For server streams, write `Ok(bytes)` to `resWritable` and close it when finished.
+Only unary and server-streaming methods support raw handlers. Other method kinds fail TypeScript checks and throw during raw registration.
+Existing typed registration helpers, middleware, and clients stay unchanged.
+
+Without middleware, River does not decode raw requests. With middleware, River decodes the request once for it and gives the handler the original bytes.
+River never re-encodes raw responses.
+
+**Never read `request.buffer`.** The request is a view into a larger buffer. Use the `Uint8Array` view itself.
```
jackyzha0: we should make this type-wise impossible if we can

protobuf/service.ts
```diff
@@ -19,6 +28,44 @@ export interface RegisteredMethod {
...
 }
 
+/** Request decoding must return a protobuf message or a raw byte view. */
+export interface MethodCodec {
+  decodeRequest(bytes: Uint8Array): unknown;
+  encodeResponse(payload: unknown): Uint8Array;
+}
+
+// Codec metadata must not change the public registration shape.
+const methodCodecs = new WeakMap<RegisteredMethod, MethodCodec>();
```
jackyzha0: hm i dont super like the module-level weakmap here, any reason we also cant reuse the existing codec interface for this?

my intuition here is that we can borrow some ideas from restate's typescript sdk and instead reason about serde interfaces per handler

https://github.com/restatedev/sdk-typescript/blob/9bb25706cee9d5c7118600251037332b676caa18/packages/libs/restate-sdk-core/src/serde_api.ts#L262

```ts
// codec/types.ts: the existing interface, made generic. Bare `Codec` is unchanged.
interface Codec<T = object, TInit = T> {
  toBuffer(obj: TInit): Uint8Array; // what a handler returns
  fromBuffer(buf: Uint8Array): T; // what a handler receives
}

// @replit/river/protobuf
serde.binary; // Codec<Uint8Array>
serde.message(HelloRequestSchema); // Codec<HelloRequest, MessageInit<HelloRequest>>

ProtoService.define(Greeter, {
  // unchanged; implicitly serde.message(input) / serde.message(output)
  sayHello: (req, ctx) => Ok({ message: `hi ${req.name}` }),

  // explicit serdes: value types come from the serdes, method kind from position
  serverStream: withSerde(
    { input: serde.binary, output: serde.binary },
    ({ request, ctx, resWritable }) => {
      resWritable.write(Ok(cachedBytes(request)));
      resWritable.close();
    },
  ),

  // mix per direction; any Codec works, not just bytes or messages
  shout: withSerde(
    { input: serde.message(HelloRequestSchema), output: textCodec },
    (req) => Ok(req.name.toUpperCase()),
  ),
});

// hoisted out of define(): pass the descriptor so the kind can be inferred
const sayHelloRaw = withSerde(
  Greeter.method.sayHello,
  { input: serde.binary, output: serde.binary },
  (bytes) => Ok(bytes),
);
```

__tests__/protobuf-raw.test.ts
Armster15: those types are for ensuring backwards compatibility with the old version's of handlers. idk if its worth keeping tests for ensuring backwards compat, made sense to me but may be unnecessary

protobuf/service.ts
Armster15: yea good call out i also agree here we shouldnt be storing anything module level weakmap if we can't. 

thanks for linking the code examples to restate's sdk! ill take a look at the broader impl patterns for a better impl of this

__tests__/protobuf-raw.test.ts
jackyzha0: imo its fine to remove

Armster15: removed

README.md
Armster15: yea good call. api is modified to ensure this (see pr desc)

protobuf/service.ts
```diff
@@ -43,14 +48,24 @@ interface ServiceConfiguration<Context extends object, State extends object> {
...
+type ErasedHandler = (...args: Array<never>) => unknown;
+
+function normalizeHandler(
+  handler:
+    | ErasedHandler
+    | { handler: ErasedHandler; raw: RegisteredMethod['raw'] },
+) {
+  return typeof handler === 'function' ? { handler, raw: undefined } : handler;
+}
+
```
Armster15, line 60: note: this is an intentional abstraction btw in order to make this more explicit.

protobuf/service.ts
Armster15: I worked through a few versions of this:

1. **Removed the WeakMap.** It was originally there to keep serialization metadata out of the public registration shape, but it made that metadata depend on object identity. Copying or reconstructing a registration could lose it. The registration now owns its raw mode directly.

2. **Implemented the generic `Codec` and serde approach.** I tried `Codec<T, TInit>`, `serde.binary`, `serde.message`, and `withSerde` with independent input/output codecs. It worked, but preserving context/state/metadata inference and preventing mismatched codec/handler combinations brought in extra overloads and branding. It also expanded the scope to arbitrary decoded values, which this use case doesn't need.

3. **Tried a callable wrapper with codec metadata.** This was an attempt to remove the branding. Putting that callable type alongside ordinary typed handlers broke contextual inference in mixed services, so I dropped that approach.

4. **Tried a smaller class-based serde wrapper.** That worked and removed some of the type machinery, but required an explicit method descriptor and still needed an annotation for context-dependent state initialization. It was simpler than the first serde implementation, but still more machinery than the raw-byte use case required.

5. **Compared that with direct raw registrations.** The current version uses `raw: 'output'` for typed requests and byte responses, or `raw: 'both'` for bytes in both directions. Existing handler inference stays intact, and one normalization helper handles the function/object distinction.

I also removed tests for the discarded serialization machinery and consolidated the remaining coverage: 604 lines became 285, with 9 focused checks instead of 30. Mutation checks confirmed they still catch re-encoding, missing input copies, lost registration modes, and broken error handling.

Reusing `Codec` is technically viable. I left it unchanged because the narrower raw-only API covers the current requirements without introducing a general serialization API.

~ written by Zerg :space_invader: ([wp-a4078f5f](https://zerg.zergrush.dev/chat?id=wp-a4078f5f))

protobuf/types.ts
```diff
@@ -136,6 +194,22 @@ export type ServiceImpl<
...
+  [MethodName in keyof Service['method']]?:
+    | ServiceImpl<Service, Context, State, ParsedMetadata>[MethodName]
+    | RawHandler<
+        Service['method'][MethodName] & DescMethod,
+        Context,
+        State,
+        ParsedMetadata
+      >;
+};
+
```
Armster15, line 206: going to clean this up

Armster15: Simplified the private raw-handler helper in f532bf1 from five type parameters to three by passing the assembled handler context. Also removed the unused error-type parameter.

Kept the mapped type here explicit so each method retains its own request type and context/state/metadata inference. Public signatures and emitted JS are unchanged. All 794 tests and checks/build pass.

~ written by Zerg :space_invader: ([wp-a4078f5f](https://zerg.zergrush.dev/chat?id=wp-a4078f5f))

protobuf/service.ts
```diff
@@ -17,6 +21,7 @@ export interface RegisteredMethod {
   readonly method: DescMethod;
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   readonly impl: MethodImpl<DescMethod, any, any, any>;
+  readonly raw?: 'both' | 'output';
```
Armster15, line 24: if youre strongly opinionated on things like string literals vs enums lmk

jackyzha0 (approved):

